PR #4, opened 2020-12-14
Fix URL in git clone command

rdnfn:
Fixed the URL in the clone command to the gtc-biodiversity one.

Related question: is necessary to use ssh for this clone command (via `git:github.com`) or could we also just let it be via https without verification, as in `git clone --recurse-submodules https://github.com/ai4er-cdt/gtc-biodiversity.git`. The latter may have less error potential.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ __Table of contents:__
 ## 1. Getting started
 ### 1.1 Cloning the project and its submodules
 To get started, first clone the github repository via:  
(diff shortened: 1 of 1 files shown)


Croydon-Brixton:
Great catch, thank you @rdnfn ! 

Since it's a public repo we could also use https for the cloning, indeed. I think only for later push and pull interactions we need to use SSH due to the project settings (only ppl from the AI4ER Orga can make changes)

Issue #19, opened 2021-02-05
Create GeoGraphViewer


rdnfn:
**Description:** create an interactive and spatial data visualisation module that allows to visualize all stages of our models from input to graph to metrics. Ideally we would also be able to include temporal information in a useful way (either via automatic transformation or slider).

**Features:**

###### Basic (minimum viable product)
- [x] Show basic visualisation of GeoGraph
- [x] Folium
- [x] ipyleaflet
- [x] Enable habitat selection
- [x] Add polygons to visualisation
  - [x] Add button selection to have graph and polygons in one line
- [x] Visualisation: Show poorly connected (single-edge) nodes #easy
- [x] Visualisation: Show individual components
- [x] Metrics: average component size
- [x] Metrics: add number of components to metrics panel
- [x] Metrics: add number of single edge nodes to metrics panel
- [ ] Diff: Enable time change via slider
- [ ] Diff: Enable diff selection (double slider) and visualisation
- [x] Diff: Show node diff
- [ ] Diff: Show polygon diff
- [ ] Allow full usage of dashboard (incl. data loading) via only voila CLI interface without need to code anything in Python (maybe with limited functionality). _This would radically broaden our user-base, since many in community appear to use R instead of Python._
- [x] Add order (so last added graph metrics are shown)

###### Extended (nice to have if time)
- [ ] Metrics
  - [x] Show standard metrics in dashboard #conventional
  - [x] Allow computation of standard metrics in current view of graph #conventional
- [ ] Add heatmap layer for some standard metrics #conventional
- [ ] Create R interface: building on the viola CLI interface
- [x] Show graph components distinctly (maybe creating a coloured buffer mostly transparent buffer around them)
- [x] Add custom style
  - [x] Add geograph logo
  - [x] Add custom dark-ish color scheme
- [ ] Allow user to draw barrier on map
- [ ] Remove edges accordingly

EDIT:
- 2 March: added extended feature list

rdnfn:
Potential libaries to build on:
- `folium` (already used for our input visualisation, see the [notebook here](https://github.com/ai4er-cdt/gtc-biodiversity/blob/exp/geopandas-chernobyl-exploration/notebooks/exploratory/rdnfn-1-geopandas.ipynb))
- `descarteslab`(with its interactive map, [see here](https://github.com/descarteslabs/descarteslabs-python/blob/master/descarteslabs/workflows/interactive/map_.py))
[...]

PR #21, opened 2021-02-12
Feature/graph-visualisation

rdnfn:
This PR adds functions to create a graph from polygon shape data, and functions to visualise that graph as used in the recent demo. This addresses part of issue #19.

@ Reviewers: Sorry for the length of the PR. It is slightly longer than intended, will try to make more regular PRs in the future. In any case, I am very open to feedback or changes! Thanks for taking the time!

**Main changes:**
- Added `graph_tools.py` module that allows for the creation of graphs from polygon data
- Added `visualisation.py` module that enables the visualisation of polygons together with a graph
- Added notebooks that demo how to use these functions together on the Chernobyl data

**Other changes:**
- Added loading functions for Polesia data (`chernobyl.py`)
- Added simple polygon test data in GeoDataFrames for fast debugging (`src/data_loading/test_data.py`)
- Added networks standard notation to good names in pylintrc
- Updated `requirements/environment.yml`
- Updated constants.py
- Update `.gitignore`
- Add other notebooks that explore Polesia data

**Things to discuss:**
- I have added dependencies to the conda `requirements/environment.yml` file instead of pip because some libraries recommend it (e.g. https://geopandas.org/install.html), but if people have tested the pip stuff on Jasmin, I am happy to transfer anything new to the pip requirements file. Note though that using pip may be less stable across platforms, i.e. external users.


pylintrc
```diff
@@ -184,7 +184,7 @@ evaluation=10.0 - ((float(5 * error + warning + refactor + convention) / stateme
 [BASIC]
 
 # Good variable names which should always be accepted, separated by a comma
-good-names=main,_
+good-names=main,_,G
```
Croydon-Brixton, line 187: G is added as a variable name for a generic graph as I understand? Could we use e.g.  `graph` instead? 

FYI on why:
One-letter function/variable names violate most styleguides (see e.g. https://www.python.org/dev/peps/pep-0008/#function-and-variable-names), especially if they are upper case. 

The rationale is normally the following ([here](https://github.com/kettanaito/naming-cheatsheet)'s a cool naming cheat-sheet, if you're interested):
One-letter variables are hard to interpret by users/readers (without context - they tend to require the user to scroll to the definition of the variable to see what its meaning is). For simple scripts like ours that's easy to do, for huge projects that can become a problem. The rationale behind the no-upper-case is that complete upper case names are reserved for constants (so a user knows they should never change this) or class definitions (e.g. Graph or DataFrame). If people strictly follow this rule then you always know what you're dealing with (a variable/function, a constant or a class) without having to check back. Ideally, variable names should be so intuitive that it would be the one of the first things you would guess.

src/data_loading/chernobyl.py
```diff
@@ -0,0 +1,31 @@
...
+    This function loads the Polesia landcover data from shared workspace on Jasmin.
+    Note that this requires access to Jasmin and the relevant shared workspaces.
+
+    Args:
+        variant (str, optional): either "biotope" or "vegetation".
+            Defaults to "biotope".
+
+    Returns:
+        gpd.GeoDataFrame: shape file data loaded as polygons in data frame
+    """
```
Croydon-Brixton, line 20: Very nice & clear documentation! 

Just an FYI and in the interest of sharing best practices 😉 

As best practice for doc-strings [PEP-257](https://www.python.org/dev/peps/pep-0257/) lists (among other):

- Triple quotes are used even though the string fits on one line. This makes it easy to later expand it.
- There's no blank line either before or after the docstring.
- The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".
- Unless the entire docstring fits on a line, place the closing quotes on a line by themselves. This way, Emacs' fill-paragraph command can be used on it.

Applied to this docstring for example, this would give:

```suggestion
    """Return Polesia landcover data from shared workspace on Jasmin.
    
    Note that this requires access to Jasmin and the relevant shared workspaces.

    Args:
        variant (str, optional): Which variant of  the data to load. Must be 
            either "biotope" or "vegetation". Defaults to "biotope".

    Returns:
        gpd.GeoDataFrame: shape file data loaded as polygons in data frame
    """
```

src/data_loading/chernobyl.py
```diff
@@ -0,0 +1,31 @@
+"""
```
Croydon-Brixton, line 1: Shall we rename this script to `utils.py` or `loading_utils.py` as we might want to add more loading routines?

src/data_loading/test_data.py
```diff
@@ -0,0 +1,115 @@
...
+
+def get_square_row(
+    num_squares: int = 2,
+    translation: Tuple[float, float] = (2, 0),
+    square_side_len: float = 1,
+    origin: Tuple[float, float] = (0, 0),
+    crs: str = UTM35N,
+) -> gpd.GeoDataFrame:
+    """
+    Creates a row of squares with side length `square_side_len`,
```
Croydon-Brixton, line 21: Nice and clear docstrings 👍  - see comment on PEP-257 in `chernobyl.py` for further best practices.

src/data_loading/test_data.py
```diff
@@ -0,0 +1,115 @@
...
+        tmp_square = shapely.affinity.translate(square, xoff=xoff, yoff=yoff)
+        out_gdf.loc[i] = [i, tmp_square, tmp_square.area]
+
+    out_gdf = out_gdf.set_crs(crs)
+
+    return out_gdf
+
+
+def get_polygon_gdf(name: str = "squares_apart") -> gpd.GeoDataFrame:
+    """Loads different test cases consisting of rows of squares.
```
Croydon-Brixton, line 66: PEP-257

```suggestion
    """Load different test cases consisting of rows of squares.
```

src/models/graph_tools.py
```diff
@@ -0,0 +1,77 @@
...
+from tqdm import tqdm
+
+
+def create_nx_graph(
+    gdf: gpd.GeoDataFrame, attributes: Optional[List[str]] = None
+) -> nx.Graph:
+    """
+    Creates networkx graph from GeoDataFrame. For each polygon in gdf a node (=vertex)
+    is created. Edges are created when the polygons of two nodes overlap or touch
+    borders. Largely taken from a notebook by @herbiebradley.
```
Croydon-Brixton, line 20: I guess this file will become obsolete when the network `GeoGraph` object will be merged? 
Shall we push this file and add a #TODO to remember to remove it when we merge with the GeoGraph branch? (Or merge with the GeoGraph branch in the first place?)

```suggestion
    Create networkx graph from GeoDataFrame.   #TODO: Refactor to GeoGraph
    
    For each polygon in gdf a node (=vertex) is created. 
    Edges are created when the polygons of two nodes overlap or touch borders. 
    Largely taken from a notebook by @herbiebradley.
```

src/models/visualisation.py
```diff
@@ -0,0 +1,213 @@
...
+
+import folium
+import geopandas as gpd
+import networkx as nx
+import shapely.geometry
+from src.constants import CHERNOBYL_COORDS_WGS84, UTM35N, CEZ_DATA_PATH
+
+
+def create_graph_visualisation(
+    m: folium.Map = None,
```
Croydon-Brixton, line 15: Minor: I'd suggest refactoring this from `m` to `map` (to stick with naming conventions and be slightly more descriptive)

```suggestion
    map: folium.Map = None,
```

src/models/visualisation.py
```diff
@@ -0,0 +1,213 @@
...
+    folium_tile_list: Optional[List[str]] = None,
+    location: Tuple[float, float] = CHERNOBYL_COORDS_WGS84,
+    crs: str = UTM35N,
+    add_layer_control: bool = False,
+) -> folium.Map:
+    """
+    This function creates a graph visualisation that puts the polygons in `polygon_gdf`
+    and the `graph` on a folium map. It is intended that the graph was build from
+    `polygon_gdf`, but it is not required. If given `m`, it will be put on this existing
+    folium map.
```
Croydon-Brixton, line 29: nit: PEP-257 

```suggestion
    Create a visualisation map of the given polygons and  `graph` in folium. 
    
    The polygons in `polygon_gdf` and `graph` are displayed on a folum map. 
    It is intended that the graph was build from `polygon_gdf`, but it is not required. 
    If given `map`, it will be put on this existing folium map.
```

src/models/visualisation.py
```diff
@@ -0,0 +1,213 @@
...
+
+    Args:
+        m (folium.Map, optional): map to add polygons and graph to. Defaults to None.
```
Croydon-Brixton, line 32:
```suggestion
        map (folium.Map, optional): map to add polygons and graph to. Defaults to None.
```

src/models/visualisation.py
```diff
@@ -0,0 +1,213 @@
...
+from typing import Optional, List, Tuple, Callable
+
+import folium
+import geopandas as gpd
+import networkx as nx
+import shapely.geometry
+from src.constants import CHERNOBYL_COORDS_WGS84, UTM35N, CEZ_DATA_PATH
+
+
+def create_graph_visualisation(
```
Croydon-Brixton, line 14: Super nice func! 💯 🥳

src/models/visualisation.py
```diff
@@ -0,0 +1,213 @@
...
+        nodes.add_to(m)
+
+    if add_layer_control:
+        folium.LayerControl().add_to(m)
+
+    return m
+
+
+def create_node_edge_geometries(
+    G: nx.Graph, crs: str = UTM35N
```
Croydon-Brixton, line 129: nit: variable naming

```suggestion
    graph: nx.Graph, crs: str = UTM35N
```

src/models/visualisation.py
```diff
@@ -0,0 +1,213 @@
...
+
+    Returns:
+        Callable[[], dict]: style function
+    """
+
+    return lambda x: {"fillColor": color, "color": color}
+
+
+def add_cez_to_map(
+    m: folium.Map,
```
Croydon-Brixton, line 179: nit: variable naming
```suggestion
    map: folium.Map,
```

Croydon-Brixton (approved):
img

👍 Looks really good to me! Thanks for writing the nice docstrings - very readable in general! 

I mainly added some optional suggestions here and there and shared some links to best practices in case you're interested.

requirements/environment.yml
```diff
@@ -4,6 +4,12 @@ channels:
   - defaults
 dependencies:
   - python=3.8
+  - jupyterlab
```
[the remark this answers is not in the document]
rdnfn: Yep, I agree! My experience with conda has been mixed as well. But since the documentation strongly recommends using conda for GeoPandas (see screenshot below from https://geopandas.org/install.html#installing-with-pip), maybe we should just install GeoPandas with conda and then everything else with pip? That should work, not sure it's more stable than just using pip. We definitely don't want anybody to need to compile C libraries themselves.

Also, has the master environment been tested from scratch? Does geopandas work with no problem there, I haven't had time to do it yet. If it does I am also happy to just keep it all pip for now.

Screenshot 2021-02-12 at 18 25 35

src/constants.py
```diff
@@ -26,3 +31,10 @@
 UTM35N = "EPSG:32635"  # https://epsg.io/32635 - preferred crs for chernobyl region
 UCS2000_TM10 = "EPSG:6384"  # https://epsg.io/6384 - reference system for ukaraine
 PREFERRED_CRS = UTM35N  # backwards compatability - preferred crs for chernobyl region
+
+# Coordinates
+CHERNOBYL_COORDS_WGS84 = (
```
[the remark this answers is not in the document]
rdnfn: Thanks for the tip!

pylintrc
rdnfn: Yeah I agree. I was sticking with the standard notation of networkx library (e.g. https://networkx.org/documentation/stable/tutorial.html) but I don't actually like this either. I think they stuck with it because G the standard maths notation for a graph. But it's python and not maths, so very happy to switch to the more readable `graph`.

@herbiebradley: you might interested to also weigh in on this as well. I think you also used the G notation, would you be happy to switch to `graph` or would you rather stick with the networkx notation?

src/data_loading/chernobyl.py
rdnfn: Brilliant, thanks for the feedback! That's a good pointer, will try to change them accordingly.

src/data_loading/chernobyl.py
rdnfn: Haha good catch, that module used to have more than one function but now with just one it doesn't make really sense anymore. My suggestion would actually be to make it a `jasmin.py` module that includes data loaders for data on our group workspace. When loading the module we can include a check if the user has access to the workspace and throw a warning if not.

src/models/visualisation.py
rdnfn: I had this originally but `map` is a built-in function (https://docs.python.org/3.8/library/functions.html#map), therefore sticking to the standard folium notation of `m` here. Could call it `fmap` or `folium_map` alternatively?

src/models/visualisation.py
rdnfn: Cheers! :)

src/models/graph_tools.py
rdnfn: Yep, I think adding the TODO is a good idea, that way we can have functioning demo version in the master now. It will be easy to replace it once the Graph class is ready.

src/models/visualisation.py
rdnfn: Thanks for the pep-257 examples, that helps :)

src/data_loading/chernobyl.py
rdnfn: Done, renamed it to `jasmin.py`. See f723bd6

rdnfn:
Thanks to @Croydon-Brixton for the detailed review, very helpful! To summarize the remaining discussion points that are still open before I merge the commit:
- Notation for folium map (m or fmap or folium_map)
- Notation for networkx graph (G or graph)
- Should we use conda to install GeoDataFrame

No need to resolve this before the weekend, have a nice weekend everyone!

pylintrc
herbiebradley: I used G because it seems like one of the few cases where one letter variable names are intuitive and readable, but happy to switch to graph if we want.

src/models/graph_tools.py
herbiebradley: Agreed, we can just merge the two files when we add branch with  `GeoGraph` to master.

src/models/visualisation.py
herbiebradley: I would favour `folium_map` since there's no real reason to use single letter variable names here.

requirements/environment.yml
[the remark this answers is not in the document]
herbiebradley: I tested the master branch installation from scratch, and it finished without errors. Doing this means all the packages including GeoPandas, fiona, etc are retrieved from the pip database, so I think the dependencies are fine with pip. I think some of the dependencies already exist in on JASMIN (e.g. `gdal-config` utility is there, which is a fiona dependency).

Therefore I think we can just have all dependencies in the `dev-requirements.txt` file.

Croydon-Brixton (approved):

src/models/visualisation.py
rdnfn: Done!

requirements/environment.yml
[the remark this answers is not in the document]
rdnfn: Great, thanks for testing it! I moved the dependencies to pip (via `dev-requirements.txt`) now.

herbiebradley (commented):
Looks good, but to be super nitpicky you might want to remove the G variable name from `pylintrc` now that we';ve decided not to use it.👍

Croydon-Brixton (approved):

herbiebradley (approved):

PR #24, opened 2021-02-22
Feature/graph analysis loading

herbiebradley:
This PR contains all the loading code for `geograph.py`. You can load the graph from:

- a raster image, either passing a path to a GeoTiff file or a numpy array containing the data. This will polygonise using a function in `rasterio_utils.py`, using `connectivity=4` and `buffer(0)`. Then the vector data will be saved to a GPKG or Shapefile and loaded into a graph.
- a GPKG or Shapefile
- a GeoPandas dataframe
- a saved graph + rtree object, saved as a pickle file or compressed pickle file (`bz2` and `gzip` are supported)

After any of these,, the graph and rtree will be saved (by default to a `bz2` file since it has the highest compression).

The `_dataframe_to_graph()` function now also supports a `tolerance` parameter, which will create edges between all polygons which are within `tolerance` metres of each other. This is done by expanding the original polygon using `.buffer(tolerance)` and checking `intersects()`.

Each node's name is the integer index it had in the dataframe. By default, the node attributes are:

- a "representative point" for the polygon (see [here](https://shapely.readthedocs.io/en/stable/manual.html#object.representative_point))
- the polygon area
- the polygon perimeter
- the class label (when converting from raster)
- the polygon object itself, under node['geometry']

Additional attributes can be added by including them in a dataframe column (when passing a dataframe or when loading from a vector file). However, I strongly recommend we always use column names of `class_label` and `geometry` for the labels and polygons to keep things simple.

This PR also contains a `merge_nodes` function which accepts a list of nodes - this will create a new node with a neighbour list and polygon attribute which is the union of the nodes in the node list, while deleting all the old nodes.

### DIFF
--- a/src/data_loading/rasterio_utils.py
+++ b/src/data_loading/rasterio_utils.py
@@ -122,12 +122,20 @@ def polygonise(
     crs: Optional[str] = None,
 ):
     """
-    Polygonise 2D numpy array containing raster data.
+    Convert 2D numpy array containing raster data into polygons.
+
+    This implementation uses rasterio.features.shapes, which uses GDALpolygonize
+    under the hood.
+
+    References:
+    (1) https://rasterio.readthedocs.io/en/latest/api/rasterio.features.html
+    (2) https://gdal.org/programs/gdal_polygonize.html
 
     Args:
         data_array (np.ndarray): 2D numpy array with the raster data.
-        mask (Optional[np.ndarray], optional): Boolean mask that can be applied
-        over the polygonisation. Defaults to None.
+        mask (Optional[np.ndarray], optional): Boolean mask that can be applied.
+        Values of False or 0 will be excluded from feature generation. Must
+        evaluate to bool.
         connectivity (int, optional): Use 4 or 8 pixel connectivity for grouping
         pixels into features. 8 can cause issues, Defaults to 4.
         apply_buffer (bool, optional): Apply shapely buffer function to the
--- a/src/data_loading/rasterio_utils.py
+++ b/src/data_loading/rasterio_utils.py
@@ -1,51 +1,56 @@
-"""A collection of utility functions for data loading with rasterio"""
+"""A collection of utility functions for data loading with rasterio."""
 
-from typing import Optional, Union, Iterable, Tuple
+from typing import Iterable, Optional, Tuple, Union
+
+import affine
+import geopandas as gpd
 import numpy as np
-from rasterio.io import DatasetReader
 from rasterio.crs import CRS
+from rasterio.features import shapes
+from rasterio.io import DatasetReader
 
 
 class CoordinateSystemError(Exception):
-    """Basic exception for coordinate system errors"""
+    """Basic exception for coordinate system errors."""
 
 
 class InvalidUseError(Exception):
-    """Basic exception for invalid usage of functions"""
+    """Basic exception for invalid usage of functions."""
 
 
 def get_thumbnail(
     data: DatasetReader,
     band_idx: Optional[int] = 1,
-    height: Optional[int] = 100,
+    height: Optional[int] = None,
     width: Optional[int] = None,
 ) -> np.ndarray:
     """
-    Calculate a thumbnail for a given band of a rasterio data
+    Calculate a thumbnail for a given band of a rasterio data.
 
     Args:
         data (DatasetReader): rasterio data handle
-        band_idx (Optional[int], optional): The band index for which to caluclate the
-            thumbnail. Defaults to 1.
-        height (Optional[int], optional): The desired height of the thumbnail.
-            If only the height is set, the width will be automatically determined
-            from the datas aspect ratio. Defaults to 100.
-        width (Optional[int], optional): The desired width of the thumbnail.
-            Defaults to None.
+        band_idx (int, optional): The band index for which to calculate the
+        thumbnail. Defaults to 1.
+        height (int, optional): The desired height of the thumbnail. If only the
+        height is set, the width will be automatically determined from the datas
+        aspect ratio. Defaults to 100.
+        width (int, optional): The desired width of the thumbnail. Defaults to
+        None.
 
     Returns:
         np.ndarray: The 2D numpy array representing the thumbnail as calculated
             from nearest neighbour resampling.
     """
-
     aspect_ratio = data.height / data.width
-
-    if height and not width:
+    if not height and not width:
+        height = 100
+    elif height and not width:
         width = int(height / aspect_ratio)
     elif width and not height:
         height = int(width * aspect_ratio)
 
-    assert height > 0 and width > 0, "Output height and/or width must be specified."
+    # Output height and/or width must be specified.
+    assert height > 0 and width > 0
 
     return data.read(band_idx, out_shape=(int(height), int(width)))
 
@@ -53,8 +58,8 @@ def get_thumbnail(
 def read_from_lat_lon(
     data: DatasetReader,
     band_idxs: Union[int, Iterable[int]],
-    lat: Tuple[float],
-    lon: Tuple[float],
+    lat: Tuple[float, float],
+    lon: Tuple[float, float],
     **kwargs,
 ) -> np.ndarray:
     """
@@ -106,3 +111,46 @@ def read_from_lat_lon(
     return data.read(
         indexes=band_idxs, window=((row_min, row_max), (col_min, col_max)), **kwargs
     )
+
+
+def polygonise(
+    data_array: np.ndarray,
+    mask: Optional[np.ndarray] = None,
+    connectivity: int = 4,
+    apply_buffer: bool = True,
+    transform: affine.Affine = affine.identity,
+    crs: Optional[str] = None,
+):
+    """
+    Polygonise 2D numpy array containing raster data.
+
+    Args:
+        data_array (np.ndarray): 2D numpy array with the raster data.
+        mask (Optional[np.ndarray], optional): Boolean mask that can be applied
+        over the polygonisation. Defaults to None.
+        connectivity (int, optional): Use 4 or 8 pixel connectivity for grouping
+        pixels into features. 8 can cause issues, Defaults to 4.
+        apply_buffer (bool, optional): Apply shapely buffer function to the
+        polygons after polygonising. This can fix issues with the polygonisation
+        creating invalid geometries.
+        transform (affine.Affine, optional): Affine transformation to apply
+        when polygonising. Defaults to the identity transform.
+        crs (Optional[str], optional): Coordinate reference system to set on the
+        resulting dataframe. Defaults to None.
+
+    Returns:
+        gpd.GeoDataFrame: GeoDataFrame containing polygon objects.
+    """
+    poly_generator = shapes(
+        data_array, mask=mask, connectivity=connectivity, transform=transform
+    )
+    results = list(
+        {"properties": {"class_label": int(val)}, "geometry": shape}
+        for shape, val in poly_generator
+    )
+    df = gpd.GeoDataFrame.from_features(results, crs=crs)
+    if apply_buffer:
+        # Redraw geometries to ensure polygons are valid.
+        df.geometry = df.geometry.buffer(0)
+
+    return df
--- a/src/data_loading/rasterio_utils.py
+++ b/src/data_loading/rasterio_utils.py
@@ -42,11 +42,11 @@ def get_thumbnail(
             from nearest neighbour resampling.
     """
     aspect_ratio = data.height / data.width
-    if not height and not width:
+    if height is None and width is None:
         height = 100
-    elif height and not width:
+    elif height is not None and width is None:
         width = int(height / aspect_ratio)
-    elif width and not height:
+    elif width is not None and height is None:
         height = int(width * aspect_ratio)
 
     # Output height and/or width must be specified.
@@ -115,11 +115,11 @@ def read_from_lat_lon(
 
 def polygonise(
     data_array: np.ndarray,
-    mask: Optional[np.ndarray] = None,
+    mask: Optional[np.ndarray],
+    transform: affine.Affine,
+    crs: Optional[str],
     connectivity: int = 4,
     apply_buffer: bool = True,
-    transform: affine.Affine = affine.identity,
-    crs: Optional[str] = None,
 ):
     """
     Convert 2D numpy array containing raster data into polygons.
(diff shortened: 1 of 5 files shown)


src/models/geograph.py
```diff
@@ -14,25 +27,92 @@ class GeoGraph:
...
+            vector_path (str or pathlib.Path, optional): A path to a file of
+            vector data, either in GPKG or Shapefile format. Defaults to None.
+            raster_load_path (str or pathlib.Path, optional): A path to a file
+            of raster data in GeoTiff format. Defaults to None.
+            raster_save_path (str or pathlib.Path, optional): A path to a file
+            to save the polygonised raster data in. A path to a GPKG file is
+            recommended, but Shapefiles also work. Defaults to None, which saves
+            to the project root in the `data` folder in GPKG format.
+            tolerance (float, optional): Adds edges between neighbours that are
+            at most `tolerance` metres apart. Defaults to 0.
```
Croydon-Brixton, line 73:
```suggestion
            at most `tolerance` units apart. Defaults to 0.
```
(since ppl might load graphs with other base units)

src/models/geograph.py
```diff
@@ -43,20 +123,179 @@ def graph(self) -> nx.Graph:
     def graph(self, new_graph):
         self._graph = new_graph
 
-    def _load_graph(self, graph_path: str) -> None:
+    @property
+    def rtree(self) -> STRtree:
```
Croydon-Brixton, line 127: Nice! This will be very handy to have as property!

src/models/geograph.py
```diff
@@ -43,20 +123,179 @@ def graph(self) -> nx.Graph:
...
+
+        Returns:
+            STRtree: The Rtree object used to build the graph.
+        """
+        if vector_path.suffix not in (".shp", ".gpkg"):
+            raise ValueError("Argument `vector_path` should be a GPKG or Shapefile.")
+        # First try to load as GeoPackage, then as Shapefile.
+        if slice is not None:
+            dataframe = gpd.read_file(
+                vector_path, rows=load_slice, enabled_drivers=["GPKG", "ESRI Shapefile"]
```
Croydon-Brixton, line 159: Out of curiosity, what does the `load_slice` do? Is that if we wanted to only load a few subelements of a dataframe?

rdnfn (approved):
Looks like some highly useful additions to me! Great to have these core functionalities GeoGraph implemented now, thanks for that! 👍 looks good to merge to me, I just added one minor optional suggestion.

src/models/geograph.py
```diff
@@ -43,20 +123,179 @@ def graph(self) -> nx.Graph:
...
+        save_path: pathlib.Path,
+        array: Optional[np.ndarray],
+        mask: Optional[np.ndarray],
+        transform: affine.Affine,
+        crs: Optional[str],
+    ) -> STRtree:
+        """
+        Load raster data, polygonise, then load graph and rtree.
+
+        The raster data should be in GeoTiff format.
```
Croydon-Brixton, line 181:
```suggestion
        The raster data should be in GeoTiff format.
        Polygonisation via `rasterio.features.shapes`, which uses `gdal_polygonize`.
        References:
        (1) https://rasterio.readthedocs.io/en/latest/api/rasterio.features.html
        (2) https://gdal.org/programs/gdal_polygonize.html
        
```

src/models/geograph.py
```diff
@@ -43,20 +123,179 @@ def graph(self) -> nx.Graph:
...
+
+        Raises:
+            ValueError: If `save_path` is not a GPKG or Shapefile, or if both
+            `raster_path` and `array` are None.
+
+        Returns:
+            STRtree: The Rtree object used to build the graph.
         """
-        Load networkx graph object from a pickle file.
+        if save_path.suffix not in (".shp", ".gpkg"):
```
Croydon-Brixton, line 205: What happens if we want to work in memory only (i.e. not write to disk)? Say for example for a small test, or a brief interactive analysis in jupyter? Is that currently supported?

src/models/geograph.py
```diff
@@ -43,20 +123,179 @@ def graph(self) -> nx.Graph:
...
+        if graph_path.suffix not in (".pickle", ".pkl", ".gz", ".bz2"):
             raise ValueError("Argument `graph_path` should be a pickle file.")
-        self.graph = nx.read_gpickle(graph_path)
+
+        if graph_path.suffix == ".bz2":
+            bz2_file = bz2.BZ2File(graph_path, "rb")
+            data = pickle.load(bz2_file)
+            bz2_file.close()
+        elif graph_path.suffix == ".gz":
+            gz_file = gzip.GzipFile(graph_path, "rb")
```
Croydon-Brixton, line 254: nit: not really needed since it's so short that the close comes immediately after, but maybe we want to use context `with ... ` here? (up to you, this is really nitpicky but just crossed my mind)

src/models/geograph.py
```diff
@@ -66,50 +305,135 @@ def _dataframe_to_graph(
...
+        Returns:
+            STRtree: The Rtree object used to build the graph.
         """
+        if tolerance < 0.0:
+            raise ValueError("`tolerance` must be greater than 0.")
         # If no attribute list is given, all
         # columns in df are used
         if attributes is None:
             attributes = df.columns.tolist()
+        elif not set(attributes).issubset(set(df.columns)):
```
Croydon-Brixton, line 326: Super neat! Haven't seen that before but I like it very much! 🥇 😍
Is the second `set(...)` needed?

src/models/geograph.py
```diff
@@ -66,50 +305,135 @@ def _dataframe_to_graph(
...
             total=len(geom),
         ):
-            # find the indexes of all polygons which touch the borders of or
-            # overlap with this one
-            neighbours = [
-                id_dict[id(nbr)]
-                for nbr in tree.query(polygon)
-                if nbr.touches(polygon) or nbr.overlaps(polygon)
-            ]
+            if tolerance > 0:
```
Croydon-Brixton, line 347: Sweet! Just out of curiosity, is this still 6x slower or was there a speed-up versus the `distance` based approach you mentioned this morning?

src/models/geograph.py
```diff
@@ -66,50 +305,135 @@ def _dataframe_to_graph(
...
             row_attributes = dict(zip(attributes, [row[attr] for attr in attributes]))
             # add each polygon as a node to the graph with all attributes
-            rep_point = row.geometry.representative_point()
-            self.graph.add_node(index, rep_point=rep_point, **row_attributes)
+            self.graph.add_node(
+                index,
+                rep_point=polygon.representative_point(),
+                area=polygon.area,
+                perimeter=polygon.length,
+                **row_attributes,
```
Croydon-Brixton, line 376: Another thought that just crossed my mind: 
- When we load from raster data, the class value will always be called `raster_val` at the moment right? (b.c. that's the default in your polygonize above). 
- When we load from a dataframe, the class value will always be whatever it's called in that df

==> Do we want to harmonise this in some way? E.g. a unified name like `node_value` or your `class_label` from below or so, such that we can later easily access this in queries over the graph?

src/models/geograph.py
```diff
@@ -66,50 +305,135 @@ def _dataframe_to_graph(
...
             # this dict maps polygon indices in df to a list
             # of neighbouring polygon indices
             graph_dict[index] = neighbours
             row = df.loc[index]
             # getting dict of column values in row
             row_attributes = dict(zip(attributes, [row[attr] for attr in attributes]))
             # add each polygon as a node to the graph with all attributes
-            rep_point = row.geometry.representative_point()
-            self.graph.add_node(index, rep_point=rep_point, **row_attributes)
+            self.graph.add_node(
```
Croydon-Brixton, line 371: Woop woop - collecting all those nodes !

Croydon-Brixton (commented):
That was a very nice read! Awesome job @herbiebradley 

I left a few minor comments/question here and there 😉 

great job

src/models/geograph.py
```diff
@@ -14,25 +27,92 @@ class GeoGraph:
...
-            graph. Defaults to None.
+            dataframe (gpd.GeoDataFrame, optional): A geopandas dataframe
+            object containing a `geometry` column with polygon data. To load
+            polygon class labels as node attributes, include these in a column.
+            Defaults to None.
+            attributes (Optional[List[str]], optional): columns of dataframe
+            that are added to nodes of graph as attributes. Defaults to None,
+            which will cause all attributes to be added.
+            graph_load_path (str or pathlib.Path, optional): A path to a pickle
+            file to load from, can be `.gz` or `.bz2`. Defaults to None.
```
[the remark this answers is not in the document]
herbiebradley: Thanks for the feedback - I added this to the class init docstring.

src/models/geograph.py
```diff
@@ -66,50 +305,135 @@ def _dataframe_to_graph(
...
+                    if nbr.intersects(new_polygon)
+                    and id_dict[id(nbr)] != id_dict[id(polygon)]
+                ]
+            else:
+                neighbours = [
+                    id_dict[id(nbr)]
+                    for nbr in tree.query(polygon)
+                    if nbr.intersects(polygon)
+                    and id_dict[id(nbr)] != id_dict[id(polygon)]
+                ]
```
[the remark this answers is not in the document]
herbiebradley: Thanks - I just realised we can speed it up slightly by switching the order of the statements in the if (prevents 1 `intersects` operation per iteration of the main loop), so I changed that.

src/models/geograph.py
herbiebradley: I tested the `distance` based approach fully and it was actually ~13x slower, but this approach is only a fraction slower, maybe it takes ~20% more time. This makes sense since the only extra computation here is the `buffer` call.

src/models/geograph.py
herbiebradley: If you check the polygonise function I actually changed `raster_val` to `class_label` - as I mentioned in the PR I would like to harmonise on `class label` - currently the only way it can be anything different is if a different named column is passed in a dataframe, but I'm not sure how to prevent that.

src/models/geograph.py
rdnfn: Same thought when I saw this: very neat!

herbiebradley: I just tested and the second actually isn't needed, so I'll remove it.

src/models/geograph.py
herbiebradley: It is not currently - I thought it best to have the default behaviour to be to save files (though if you do it repeatedly it just overwrites the same file). Should I add in an extra boolean argument in the init that turns off all saving?

src/models/geograph.py
herbiebradley: Yes, you can pass a python slice object using the `slice()` constructor https://www.programiz.com/python-programming/methods/built-in/slice

src/models/geograph.py
```diff
@@ -14,25 +27,92 @@ class GeoGraph:
...
+            crs (str, optional): Coordinate reference system to set on the
+            resulting dataframe. Defaults to None.
         """
         super().__init__()
         self._graph = nx.Graph()
-        if graph_path is not None:
-            self._load_graph(graph_path)
-        else:
-            self._dataframe_to_graph(dataframe)
+        if graph_save_path is None:
```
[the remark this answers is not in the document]
herbiebradley: This seems like a good idea to simplify the interface, so I'll see if I can implement it tomorrow.

src/models/geograph.py
Croydon-Brixton: Cool, thank you for the extra info! Great tests!

src/models/geograph.py
herbiebradley: I originally had issues with using `with` for some reason, but I tried it again and it seems to work so I've changed it.

src/models/geograph.py
herbiebradley: Done.

src/data_loading/rasterio_utils.py
```diff
@@ -106,3 +111,46 @@ def read_from_lat_lon(
...
+    apply_buffer: bool = True,
+    transform: affine.Affine = affine.identity,
+    crs: Optional[str] = None,
+):
+    """
+    Polygonise 2D numpy array containing raster data.
+
+    Args:
+        data_array (np.ndarray): 2D numpy array with the raster data.
+        mask (Optional[np.ndarray], optional): Boolean mask that can be applied
```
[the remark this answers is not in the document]
herbiebradley: Done.

src/data_loading/rasterio_utils.py
```diff
@@ -106,3 +111,46 @@ def read_from_lat_lon(
...
+def polygonise(
+    data_array: np.ndarray,
+    mask: Optional[np.ndarray] = None,
+    connectivity: int = 4,
+    apply_buffer: bool = True,
+    transform: affine.Affine = affine.identity,
+    crs: Optional[str] = None,
+):
+    """
+    Polygonise 2D numpy array containing raster data.
```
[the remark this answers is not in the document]
herbiebradley: Done.

src/models/geograph.py
herbiebradley: Done.

src/models/geograph.py
herbiebradley: If they're passing in a dataframe then they might as well just rename it to `class_label` anyway, right? Maybe we should just throw an error if it doesn't include `class_label`.

Croydon-Brixton: Agreed, that would simplify things a lot

herbiebradley: @Croydon-Brixton I added this as a check to the `_dataframe_to_graph()` function, but I realised that if we load in data from a vector file, such as a shapefile, this will automatically trigger the check since the resulting dataframe will not have a class label column (or it will be named something different).

Therefore I have added a new function in `vector_utils.py` which loads in any vector data, renames a specified column to "class_label", saves it in the GPKG format and returns the dataframe.

src/models/geograph.py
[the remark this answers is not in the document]
herbiebradley: @Croydon-Brixton I added this feature in the latest commit, if you think it's all good then I'll merge this PR.

src/models/geograph.py
herbiebradley: Update: I turned off saving by default so if you don't pass any save paths then it doesn't do anything.

Croydon-Brixton (approved):
@herbiebradley Looks very sweet! 
(I only added a few extra lines of documentation)

Good to merge if you ask me! 🥳 🎈

Issue #25, opened 2021-02-24
Node identification and operations


Croydon-Brixton:
This issue concerns all issues related to:

- node identification
- operating on nodes (merging / adding /removing and the corresponding updates in the underlying datastructure of `GeoGraph`, such as updating ids, areas or the rtree for example).

Croydon-Brixton:
With the new geograph object now ready, I've implemented a vanilla version of node identification with the new RTree now. We can now identify nodes based on their relative positions. 

**(C) Nodes are currently identified if:**
1.  they have the same class label
2.  they intersect/touch in some way (with possibility to disallow touching at the corners only)

**(L) Current limitations:**
1. if nodes do not touch/intersect at all, they will not be identified. This is the desired behavior for most use cases on the dense graph as I understand. For sparse subgraphs we might want the option to search the nearest candidate of the same class and add that one if it's "reasonably similar" and "closer than X tolerance" (except for maybe what Toby wanted with his identification of trees)
2. If we check for corners (when disallowing corners only) we incur a slight slow-down (I'm currently doing a [DE-9IM](https://en.wikipedia.org/wiki/DE-9IM) pattern match computation for that). Allowing corners is about 50% faster in the cases where corner-tests actually need to be performed (otherwise no notable difference).
3. The identification (allowing corners) takes about ~140 us on sci1 for one node in my test cases. This means we can identify about 3000 nodes per second
4. From my line-profiling investigations, it seems that the largest bottleneck is actually querying the RTree (takes up over half the time). Future speed-ups might be gained (in order of importance) by speeding up the rtree query, by directly leveraging GDAL implementations of the DE-9IM pattern computations and by vectorising the operations.

**Images** :
Facecolors = class labels
Blue edge = selected node for which to find identified nodes in `other_graph`
Green edge = nodes that are identified with the blue node
![image](https://user-images.githubusercontent.com/19754061/109002011-8fd55d00-769d-11eb-8d9f-882eb5e65130.png)
![image](https://user-images.githubusercontent.com/19754061/109002079-a11e6980-769d-11eb-9e30-5b816e8599b9.png)
![image](https://user-images.githubusercontent.com/19754061/109002142-b8f5ed80-769d-11eb-98ec-732fc49b028d.png)
![image](https://user-images.githubusercontent.com/19754061/109002153-bc897480-769d-11eb-9995-fc72ad48c803.png)
[...]

PR #28, opened 2021-02-24
Feature/graph analysis identification

Croydon-Brixton:
Still WIP, but I'm creating a draft PR so that people can already make early comments and check out the work for their branches.

More detailed info & finalized PR coming soon.
This addresses #25 

Note: 
There's lots of files because I also pushed the test-cases for testing node identification (the `.gpkg` files) - simply ignore these.
Note also that the branch contains merges of @herbiebradley latest code on `feature/graph-analysis-habitat` for fixing the `rtree`.  You can also ignore those.


Croydon-Brixton:
This PR consists of the `identify_nodes` function as well as a couple of test cases - closes #25 

We identify nodes based on the following tactic:
1. For each node get all candidates via an RTree query
2. Remove all nodes which have a different class label (only labels with same class will be identified with each other)
3. Compare the geometries of the nodes (3 different modes possible, which are successively more stringent: `corner`, `edge`, `interior`. `corner` identifies all nodes that have at least a 0D overlap (point, line, area), `edge` identifies if there's at least a 1D overlap (line or area), `interior` identifies if there's at least a 2D overlap)

The relevant files are:
- `src/models/binary_graph_operations.py` and `src/models/polygon_utils.py` contain the main functionality. 
- `notebooks`: include sample test cases and timings (indices `8, 9, 10`). Best to watch these on [ReviewNB](https://app.reviewnb.com/ai4er-cdt/gtc-biodiversity/pull/28)
- `src/tests` includes two scripts `create_test_data` (which creates 15 binary test cases of tiny landscapes (c.f. below) and `test_utils` which are used to plot node identification examples in the notebooks.

Other contents of this PR:
- 15 binary test cases of tiny landscapes with different nodes and class labels for doing tests of the node identification (let me know if you think I should exclude them from the PR)
- Minor fixes of linting
- Adding some defaults
- Named our index (just for fun 😄 )


Extra notes:
**Note 1:  Geopandas access**
Takeaway: Do not use `.iloc` or `.loc` indexing if you need performance. They can be 10-100x slower than directly accessing the underlying numpy array. This [stackexchange ](https://stackoverflow.com/questions/44960614/whats-the-fastest-way-to-acces-a-pandas-dataframe) issue reports the same observation.

image

src/models/binary_graph_operations.py
```diff
@@ -0,0 +1,118 @@
...
+    # Extract relevant node elements for legibility
+    poly = node["geometry"]
+    label = node["class_label"]
+
+    # Get potential candidates for overlap
+    candidate_ids = other_graph.rtree.query(poly, sort=True)
+    # Filter candidates according to the same class label
+    # fmt: off
+    candidate_ids = candidate_ids[
+        other_graph._class_label(candidate_ids) == label  # pylint: disable=protected-access
```
Croydon-Brixton, line 62: Feature:
Just noticed this now. It'd be useful to also allow identifying nodes with other (custom) classes where we determine the class-to-class mapping. I will add this as an extra small feature.

src/tests/utils.py
```diff
@@ -0,0 +1,146 @@
...
+        transform=get_array_transform(sub_array, xoff=x_lims[0], yoff=y_lims[0]),
+    )
+
+
+def plot_identified_nodes(
+    node: dict, other_graph: GeoGraph, identified_nodes: Iterable[int]
+) -> None:
+    """
+    Plot nodes that identify with `node` in `other_graph`
+
```
rdnfn, line 113: Great to have a first start on plotting, more to come hopefully 😄

src/models/binary_graph_operations.py
```diff
@@ -0,0 +1,118 @@
...
+            identifications
+        mode (str, optional): Must be one of `corner`, `edge` or `interior`. Defaults
+            to "corner".
+            The different modes correspond to different rules for identification:
+            - corner: Polygons of the same `class_label` which overlap, touch in their
+                edges or corners will be identified with each other. (fastest)
+            - edge: Polygons of the same `class_label` which overlap or touch in their
+                edges will be identified with each other.
+            - interior: Polygons of the same `class_label` which overlap will be
+                identified with each other. Touching corners or edges are not counted.
```
rdnfn, line 43: Question out of interest: are there examples where anything apart from using the `interior` mode would make sense in our case? Maybe, noise and inaccuracy in CRS projection could warrant the others? @Croydon-Brixton: you probably thought more about this, would be curious to hear what examples you were thinking about.

src/models/binary_graph_operations.py
```diff
@@ -0,0 +1,118 @@
...
+    Returns:
+        np.ndarray: List of node ids in `other_graph` which identify with `node`.
+    """
+
+    # Mode switch
+    assert mode in ["corner", "edge", "interior"]
+    have_valid_overlap = _BULK_SPATIAL_IDENTIFICATION_FUNCTION[mode]
+
+    # Extract relevant node elements for legibility
+    poly = node["geometry"]
```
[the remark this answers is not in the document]
rdnfn: Just saw now that you (rightfully) removed `graph_tools.py`, so nevermind that part of my comment.

rdnfn (approved):
Looks great to me! Just added some very minor comments, some just to start a discussion about unifying our notation & style, so those definitely don't need to be fully addressed for this to be merged.

In general excellent to see the progress here, excited for how this work can be used in the `GeoGraphTimeline`class and then consequently in our visualisation 🚀

src/models/geograph.py
```diff
@@ -589,3 +601,8 @@ def get_graph_components(
         geom = [self.df.geometry.iloc[list(comp)].unary_union for comp in components]
         gdf = gpd.GeoDataFrame({"geometry": geom}, crs=self.df.crs)
         return gdf, components
+
+    def identify_node(
+        self, node_id: int, other_graph: GeoGraph, mode: str
+    ) -> List[int]:
+        return identify_node(self.df.iloc[node_id], other_graph=other_graph, mode=mode)
```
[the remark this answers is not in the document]
herbiebradley: I know that standard conventions are to import modules https://google.github.io/styleguide/pyguide.html#22-imports

src/models/geograph.py
```diff
@@ -193,6 +195,14 @@ def crs(self):
...
         return self.df.crs
 
+    def _class_label(self, node_ids: Sequence[int]):
+        """Return class label of `node_ids` directly from underlying numpy array"""
+        return self.df.class_label.values[node_ids]
+
+    def _geometry(self, node_ids: Sequence[int]):
+        """Return geometry of `node_ids` directly from underlying numpy array"""
+        return self.df.geometry.values[node_ids]
+
```
herbiebradley, line 205: What's the anticipated use for these? I'm not sure this really saves code reuse - in fact I feel like most commonly we want to get the full `.values` array. Maybe we could make these a class property so:

```python
@property
def geometry(self):
    return self.df.geometry.values
```
which makes some things a bit neater for calling these outside the class and allows for whatever slicing you want?

src/models/binary_graph_operations.py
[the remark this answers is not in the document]
Croydon-Brixton: @rdnfn @herbiebradley 
Great point about harmonizing our notation. Which do you prefer?
- `polygon`
- `poly`
- `pgon`
- `...?`

I will then change all accordingly 😄

src/models/binary_graph_operations.py
Croydon-Brixton: Good question! 

My thoughts at the moment are:
- `interior` makes most sense when we stack graphs on top of each other along the time-axis.
- `edge` makes most sense when we want to combine two graphs which overlap/touch in a certain region to one larger graph (I was thinking scalablility here, maybe if we have to distribute different graphs on different workers on a cluster or simply if we have run an individual analysis in region A and one in region B before and want an easier way to get A+B than rerunning it on the A+B composite image by simply merging the graphs spatially). It would essentially catch "touching" polygons like in this image.
image
- `corner`: I am not yet sure in which use case we would want to do that. Potentially when working on very high-res data and small features, where maybe a river is only diagonally connected in some cases this could be useful to combine the  individual river nodes into one larger river node, despite them lying diagonally, by going over all river-type nodes and corner-identifying them with their neighbors. This could reduce the 1-pixel noise artefacts that come from rasterization.

src/models/binary_graph_operations.py
```diff
@@ -0,0 +1,118 @@
...
+            poly,
+            other_graph._geometry(candidate_ids),  # pylint: disable=protected-access
+        )
+    ]
+
+    return candidate_ids
+
+
+### Deprecated but kept for tests and backward compatibility
+def identify_node_old(
```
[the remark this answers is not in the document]
Croydon-Brixton: Good question! The difference is:
`old`: works on the data in a for loop, node-by-node (much slower)
`new`: works on the data in vectorized form, bulk operations on all nodes at once

src/models/binary_graph_operations.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # Build list of nodes in `other_graph` which identify with `node`
+    identifies_with = []
+    for candidate_id in other_graph.rtree.intersection(node["geometry"].bounds):
+
+        candidate_node = other_graph.df.iloc[candidate_id]
+        have_same_class_label = node["class_label"] == candidate_node["class_label"]
+
+        if have_same_class_label and have_valid_overlap(
+            node["geometry"], candidate_node["geometry"]
```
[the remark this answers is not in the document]
Croydon-Brixton: Ah you mean using `node['geometry']` instead of `poly`? 
Sure, if it's that I can easily change that 😉 If it's not, I'm not sure I fully understand what you mean by 
> what `poly` is or check 
?

src/models/binary_graph_operations.py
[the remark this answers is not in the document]
herbiebradley: I quite like `polygon`, and I think basically all of the code in `GeoGraph` uses that.

src/models/binary_graph_operations.py
herbiebradley: Good ideas - but do you know if the rasterisation ever produces setups like in your `edge` diagram? If there are two pixels of the same class touching each other on a full pixel edge, it should just merge them right?

src/models/geograph.py
```diff
@@ -52,7 +54,7 @@ class GeoGraph:
     def __init__(
         self,
         data,
-        crs: Union[str, pyproj.CRS],
+        crs: Union[str, pyproj.CRS] = None,
```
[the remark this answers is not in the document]
Croydon-Brixton: Interesting question,
the probably most extensive answer is the [class definition here](https://pyproj4.github.io/pyproj/dev/api/crs/crs.html), but that's likely not what you mean by the question I guess.

The main reason for allowing `str` and `pyproj.CRS` here is that this is the underlying way in which `geopandas` stores and can accept CRS information (c.f. [here](https://geopandas.org/docs/user_guide/projections.html) - if you give a string that's essentially parsed into a `pyproj.CRS` by geopandas under the hood). 
Strings for CRS are useful because they allow you to use generally defined CRS (like the UTM ones, or WGS84, ...). But if you were to do your custom projection of the sphere you would get your own CRS which does not have a canonical string name. For such use cases, where users work with their custom projections, I wanted to also allow `pyproj.CRS`. In terms of handling everything it doesn't change much, as geopandas does all of the CRS stuff for us 😄 

[Pyproj](https://pyproj4.github.io/pyproj/dev/index.html) is  the canonical CRS handling library in python that uses the C++ library [PROJ](https://proj.org/about.html) under the hood. `Rasterio` uses it's [own **binder](https://github.com/mapbox/rasterio/blob/master/rasterio/crs.py) to the PROJ library, rather than pyproj. Since they both go to the underlying PROJ tough, they interface nicely as far as I know. 

As for `ipyleaflets` I'm less familiar, they seem to be using [dictionaries](https://github.com/jupyter-widgets/ipyleaflet/blob/38a7e9d8284aa1c3d7fc465415569e312a387be6/ipyleaflet/leaflet.py#L1970) (see also [this sample notebook with custom projection](https://github.com/jupyter-widgets/ipyleaflet/blob/38a7e9d8284aa1c3d7fc465415569e312a387be6/examples/CustomProjections.ipynb)). They don't include `pyproj` in their dependencies, so they likely have their own binder to PROJ as well (as rasterio does). I would be surprised if they didn't use PROJ though, since this seems to be the defacto standard since ~30 yrs in the GIS community. 

So bottom line to your question: I think deep under the hood they are all the same (PROJ), but the interfaces may vary.

src/models/binary_graph_operations.py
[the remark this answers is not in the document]
Croydon-Brixton: Awesome - I'll change to `polygon` then 😉

src/models/geograph.py
Croydon-Brixton: Great idea acutally! 
The original use was really just to save me typing `self.df.geometry.values` all the time for readibility. But I like your idea much better - will push it with the next commit - thank you for the good suggestion!

src/models/geograph.py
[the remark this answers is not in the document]
Croydon-Brixton: Thanks for pointing these out! I actually wasn't aware - great to know. I'm adapting to importing modules in the next commit.

src/models/binary_graph_operations.py
Croydon-Brixton: Correct, in the `edge` case the polygonisation will merge this automatically, so `edge` is really only needed when users cut the raster before, polygonise them separately and then want to merge them back again.

So, in summary, the three main applications I see at the moment are:
- `interior`: for time-stacking
- `edge`: for glueing together different graphs (created by separate polygonisation
- `corner`: for fixing diagonally connected nodes

src/models/binary_graph_operations.py
```diff
@@ -0,0 +1,118 @@
...
+    ]
+    # Filter candidates accroding to correct spatial overlap
+    # fmt: on
+    candidate_ids = candidate_ids[
+        have_valid_overlap(
+            poly,
+            other_graph._geometry(candidate_ids),  # pylint: disable=protected-access
+        )
+    ]
+
```
herbiebradley, line 72: Intuitively it should be slightly faster to combine these two filters into one, right?

src/models/polygon_utils.py
```diff
@@ -0,0 +1,143 @@
...
+#  They are not guaranteed to work on lower dimensional objects (points/lines)
+CORNER_ONLY_PATTERN = "FF*F0****"
+EDGE_ONLY_PATTERN = "FF*F1****"
+# Note: since we deal with polygons only, we can use a simplified overlap pattern:
+#  If polygons overlap with more than just their edge, they will automatically overlap
+#  with their interiors
+OVERLAP_PATTERN = "T********"
+
+
+def de9im_match(pattern: str, target_pattern: str) -> bool:
```
herbiebradley, line 16: I feel like there has to be a way to make this much more efficient, maybe through stuff like `numba` or Cython, but that's something to leave for later.

herbiebradley: Side note: when searching for what the relate function does I came to this: https://github.com/libgeos/geos/blob/186bbd32fbf07d8b5d419cdfd64c14e2270a418a/src/geom/IntersectionMatrix.cpp#L89 which I think are the functions which a lot of Shapely stuff calls.

herbiebradley (approved):
Looks good, I added all the comments I could think of ;)

src/models/binary_graph_operations.py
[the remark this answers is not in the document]
rdnfn: Yeah, just gave this feedback because you specifically wrote in a comment it's for legibility. No need to change it really, it's fine either way.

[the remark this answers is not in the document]
Croydon-Brixton: Changed anyways 😉

src/models/polygon_utils.py
```diff
@@ -0,0 +1,143 @@
...
+from numpy import ndarray
+from shapely.geometry.polygon import Polygon
+from geopandas.geoseries import GeometryArray
+
+# Note: All DE-9IM patterns below are streamlined to work well with polygons.
+#  They are not guaranteed to work on lower dimensional objects (points/lines)
+CORNER_ONLY_PATTERN = "FF*F0****"
+EDGE_ONLY_PATTERN = "FF*F1****"
+# Note: since we deal with polygons only, we can use a simplified overlap pattern:
+#  If polygons overlap with more than just their edge, they will autmoatically overlap
```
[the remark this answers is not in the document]
Croydon-Brixton: Accepted

src/models/geograph.py
Croydon-Brixton: Done.

src/models/polygon_utils.py
Croydon-Brixton: Ah nice, I didn't see the clibrary implementation - yes there's definitely a way to make it faster!

Overall, I haven't concentrated on this part much bc the string comparison is much faster than the overlap computations and rtree query so the overall speedup in the function would be only around 5%. 


But I agree that we should note this for later 👍 --> I'll convert it to an issue

herbiebradley: Ah ok, I wouldn't have guessed it was such a small part of the overall time just from the code. Probably not a priority then.

src/models/binary_graph_operations.py
Croydon-Brixton: Yes, I also think that if numpy does short-circuit evaluation on these things this should be faster.

Do you know how numpy handles these type of cases?
Concretely:
Case `select_from_array[np.logical_or(condition_array1, condition_array2)]`
Does it first evaluate both `condition_array1` and `condition_array2` in the slice `[ ... ]` and then `or` the conditions (in which case it'd probably be slower bc we would calculate the geometry overlaps for shapes which won't agree in class label).
Or does it calculate the first element of `condition_array1` and then short-circuit decide if that element of `condition_array2` even needs to be calculated? (in which case I think it should be slightly faster)

src/models/polygon_utils.py
```diff
@@ -0,0 +1,143 @@
...
+    Returns:
+        np.array: Boolean array with value True, iff `poly` and the polygon in
+            `poly_array` at the given location intersect.
+    """
+    return poly_array.intersects(poly)
+
+
+def connect_with_interior_or_edge_bulk(
+    poly: Polygon, poly_array: GeometryArray
+) -> ndarray:
```
herbiebradley, line 109: I noticed we have ndarray type hints here when the actual return type is List[bool]? Which do we want for the binary_graph_operations function?

Croydon-Brixton: Good catch! Thank you - corrected it to `List[bool]` now. Either work as it's used as boolean index for slicing

src/models/binary_graph_operations.py
herbiebradley: I'm not sure about `np.logical_or` but if you use base python `or` and a generator to select then it should short-circuit and may be faster - there are some interesting timing results here on selecting from boolean arrays: https://stackoverflow.com/questions/58422690/filtering-a-numpy-array-what-is-the-best-approach

src/models/geograph.py
```diff
@@ -586,6 +622,13 @@ def get_graph_components(
...
         """
         components: List[set] = list(nx.connected_components(graph))
-        geom = [self.df.geometry.iloc[list(comp)].unary_union for comp in components]
+        geom = [self.df.geometry.loc[comp].unary_union for comp in components]
         gdf = gpd.GeoDataFrame({"geometry": geom}, crs=self.df.crs)
         return gdf, components
+
+    def identify_node(
+        self, node_id: int, other_graph: GeoGraph, mode: str
+    ) -> List[int]:
```
herbiebradley, line 631: This function currently returns an ndarray, do we want to convert it to list or change the type hint?

herbiebradley (approved):
I fixed the indexing issue discussed on Slack, and added a minor performance improvement to `add_habitat`, chmod 664 to all saved files, and some minor bugfixes. Looks good to merge now once the return type of `identify_nodes` is resolved.

src/models/geograph.py
herbiebradley: OK, fixed in 29beff4

PR #30, opened 2021-02-25
Feature/graph-analysis-visualisation

rdnfn:
This tiny PR adds basic GeoGraphViewer functionality.

It currently is only built on the prior folium plotting functions, next step will be to redo it with ipyleaflet. I am planning to maintain the basic folium support because folium can be used on some platforms that don't support ipywidgets (and therefore ipyleaflet), like google colab. Also: plotting of polygons is not part of it yet and will be added once the df attribute is merged.

**Main Changes**
- Added GeoGraphViewer class

**Other Changes**
- fixed changed name for represenative point (to rep_point)
- removed choropleth color legend because not useful in our case.

**Demo of usage:**
Screenshot 2021-02-25 at 10 48 27

### DIFF
--- a/src/models/visualisation.py
+++ b/src/models/visualisation.py
@@ -1,7 +1,7 @@
 """
-This module contains visualisation functions for vector, raster and graph data.
+This module contains visualisation functions for GeoGraphs.
 """
-
+from __future__ import annotations
 from typing import Optional, List, Tuple, Callable
 
 import folium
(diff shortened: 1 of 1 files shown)


src/models/visualisation.py
```diff
@@ -1,17 +1,72 @@
...
+    def _repr_html_(self) -> str:
+        """Return raw html of widget as string.
+
+        This method gets called by IPython.display.display().
+        """
+
+        if self.widget is not None:
+            return self.widget._repr_html_()  # pylint: disable=protected-access
+
+    def add_graph(self, graph: geograph.GeoGraph) -> None:
```
Croydon-Brixton, line 43: Wrt the comment on polymorphism above:
Since you will have a coulpe of similar functionalities like `add_graph` , `_repr_html_` or `add_layer_control` I think polymorphism could really make sense and simply things here (e.g. you wouldn't need to do the `if self.map_type == XYZ` checks all the time and the code would be more easily seperable in Folium and IpyWidgets.

If you wanna go this route and want to briefly brush up on polymorphism & inheritance in python, I recommend reading [this article](https://alan-turing-institute.github.io/rsd-engineeringcourse/ch05construction/08objects.html) first. I guess the software architecture pattern to take inspiration from in this case would be a Strategy-type pattern (see [here](https://www.tutorialspoint.com/design_pattern/strategy_pattern.htm) or [here](https://alan-turing-institute.github.io/rsd-engineeringcourse/ch05construction/09patterns.html) for more info).
I think that could might make things easier further down the line.

Croydon-Brixton (approved):
Nice! Looks good to me, only one suggestion in terms of software architecture if we plan to support both folium and ipywidgets in the future: Maybe use polymorphism & inheritance to simplify things and create clean interfaces (e.g. strategy type pattern) instead of the current mode-switch in one class which encompasses both.

More detail in the comments - let me know what you think 😉

herbiebradley (approved):
Looks good! I agree that splitting it into two classes with inheritance would make some things a bit neater, but that can come in another PR if you want to leave it until you implement the `ipyleaflet` viewer.

Croydon-Brixton:
> ... but that can come in another PR if you want to leave it until you implement the `ipyleaflet` viewer.

Agree with that 😉

src/models/visualisation.py
rdnfn: Yep definitely agree. I was thinking that it's starting to look a bit unwieldy, so polymorphism will definitely be the way to go. Thanks for the resources, very useful 👍

rdnfn:
@Croydon-Brixton @herbiebradley: Thanks for the reviewing! For the ipyleaflet version I will keep the polymorphism in mind 🚀

PR #31, opened 2021-02-25
Feature/graph analysis add_habitat function and get_graph_components function

### LINKED ISSUE #22: Allow minimal distance-based edges in a habitat graph
**Description:** Currently graph edges are only added to the graph when two polygons overlap or touch. For habitats there will almost always be a certain distance a species is able to travel across land that is not part of the habitat. Therefore we want to be able to add edges between polygons if they are within a certain _maximum travel distance_.

**Potential solution:** One way would be to add an edge to all polygons within a certain range (e.g. 10m) in the initial creation of the (parent-)graph and add an edge-attribute of `min_distance` between two polygons. Then when we add habitat subgraphs later we could easily filter the edges such that their `min_distance` is withing the `max_travel_distance`.

**Further notes:**
The `create_habitat` method of the `GeoGraph` class might take the following arguments

- `name` (e.g. `"eagle habitat"`)
- list of landcover classes that are part of the habitat (e.g. [`"pine forests", ...]`)
- `max_travel_distance`: maximum travel distance outside habitat (e.g. 50m)
- list of landcover classes that are uncrossable/impassable (e.g.` ["motorway",...]`)

### LINKED ISSUE #27: Make graph components accessible
**Description:** for the UI it would be great to have the following method/function, either as part of `GeoGraph` or as a standalone function:

```python
def get_graph_components(graph: nx.Graph()) -> gpd.GeoDataFrame():
    """Return a GeoDataFrame with graph components.
    
    This method takes an nx.Graph and determines the individual disconnected graph
    components that make up the graph. Each row of the returned GeoDataFrame 
    corresponds to a graph component, with entries in column 'geometry' being the union
    of all individual polygons making up a component.
    
    This method allows for the UI to visualise components and output their number as 
    a metric.
    
    More info on the definition of graph components can be found here:
    https://en.wikipedia.org/wiki/Component_(graph_theory)

    Args:
        graph (nx.Graph): nx.Graph of a GeoGraph
    """

```

Using the returned GeoDataFrame, the UI can then color code the components and output their number as a metric.

herbiebradley:
This PR consists of the `add_habitat` function and the `get_graph_components` components function - closes #22, closes #27 

The `add_habitat` function takes in a name, a *set* of valid landcover classes, and a `max_travel_distance`, and adds an entry in the dictionary class attribute `habitats` with the resulting habitats graph. This graph is a subgraph of the main graph consisting of only the nodes with a `"class_label"` in the habitat, and edges between all nodes less than `max_travel_distance` apart.

The `get_graph_components` function takes in an `nx.Graph` object and returns a tuple of a `GeoDataFrame` and a `List[set]` containing the connected components of the graph. The List just contains sets containing the node indices of the nodes in each component, and the dataframe just contains one row for each component, with a single `"geometry"` column that stores the union polygon for the component. The caveat with this function is that the graph object passed must be a subgraph of the main class's graph, since it relies on the indices in the dataframe class attribute.

- Polygons are now stored in the dataframe, which is available as a class property along with the crs. The dataframe is now saved with the graph and rtree when saving to a pickle file.
- Switch to the `rtree` library, which has a much richer rtree implementation than Shapely's version.
- New node attributes: the `"class_label"` and `"bounds"` for each polygon. This is so that we can do `graph.nodes(data="bounds")` to iterate over all polygon bounds quickly. Other optional node attributes are no longer supported, since they are in the dataframe anyway which is now kept.
- Bugfixes in the `merge_nodes` function.

### DIFF
--- a/src/models/geograph.py
+++ b/src/models/geograph.py
@@ -50,7 +50,7 @@ class GeoGraph:
     def __init__(
         self,
         data,
-        crs: str,
+        crs: Union[str, pyproj.CRS], 
         graph_save_path: Optional[Union[str, os.PathLike]] = None,
         raster_save_path: Optional[Union[str, os.PathLike]] = None,
         columns_to_rename: Optional[Dict[str, str]] = None,
@@ -114,7 +114,7 @@ class GeoGraph:
         super().__init__()
         self.graph = nx.Graph()
         self._habitats: Dict[str, Habitat] = {}
-        self._crs: str = crs
+        self._crs: Union[str, pyproj.CRS] = crs
         self._columns_to_rename: Optional[Dict[str, str]] = columns_to_rename
         self._tolerance: float = tolerance
 
--- a/src/models/geograph.py
+++ b/src/models/geograph.py
@@ -496,6 +496,7 @@ class GeoGraph:
         """
         if max_travel_distance < 0:
             raise ValueError("`max_travel_distance` must be greater than 0.")
+        valid_classes_set = set(valid_classes)
         hgraph: nx.Graph = deepcopy(self.graph)
         # Remove all edges in the graph, then at the end we only have edges
         # between nodes less than `max_travel_distance` apart
@@ -513,7 +514,7 @@ class GeoGraph:
         for node in tqdm(
             hgraph.nodes, desc="Generating habitat graph", total=len(hgraph)
         ):
-            if hgraph.nodes[node]["class_label"] not in valid_classes:
+            if hgraph.nodes[node]["class_label"] not in valid_classes_set:
                 invalid_indexes.append(node)
                 continue
 
@@ -525,7 +526,7 @@ class GeoGraph:
             # Query rtree for all polygons within `max_travel_distance` of the original
             for nbr in self.rtree.intersection(buff_poly.bounds):
                 # If a node is not a habitat class node, don't add the edge
-                if hgraph.nodes[nbr]["class_label"] not in valid_classes:
+                if hgraph.nodes[nbr]["class_label"] not in valid_classes_set:
                     continue
                 # Otherwise add the edge with distance attribute
                 nbr_polygon = polygons[nbr]
--- a/src/models/geograph.py
+++ b/src/models/geograph.py
@@ -442,7 +442,7 @@ class GeoGraph:
             # add each polygon as a node to the graph with useful attributes
             self.graph.add_node(
                 index,
-                rep_point=polygon.representative_point().coords[0],
+                rep_point=polygon.representative_point(),
                 area=polygon.area,
                 perimeter=polygon.length,
                 class_label=class_labels[index],
--- a/src/models/geograph.py
+++ b/src/models/geograph.py
@@ -18,7 +18,7 @@ import numpy as np
 import rasterio
 import rtree as rtree_lib
 import shapely
-from shapely.ops import unary_union
+from src.constants import PREFERRED_CRS
 from src.data_loading.rasterio_utils import polygonise
 from tqdm import tqdm
 
@@ -46,6 +46,7 @@ class GeoGraph:
         graph_save_path: Optional[Union[str, os.PathLike]] = None,
         raster_save_path: Optional[Union[str, os.PathLike]] = None,
         tolerance: float = 0.0,
+        crs: str = PREFERRED_CRS,
         **kwargs,
     ) -> None:
         """
@@ -86,13 +87,13 @@ class GeoGraph:
             not save the polygonised data.
             tolerance (float, optional): Adds edges between neighbours that are
             at most `tolerance` units apart. Defaults to 0.
+            crs (str, optional): Coordinate reference system to set on the
+            resulting dataframe. Defaults to None.
 
             **mask (np.ndarray, optional): Boolean mask that can be applied over
             the polygonisation. Defaults to None.
             **transform (affine.Affine, optional): Affine transformation to
             apply when polygonising. Defaults to the identity transform.
-            **crs (str, optional): Coordinate reference system to set on the
-            resulting dataframe. Defaults to None.
             **connectivity (int, optional): Use 4 or 8 pixel connectivity for
             grouping pixels into features. 8 can cause issues, Defaults to 4.
             **apply_buffer (bool, optional): Apply shapely buffer function to
@@ -163,6 +164,11 @@ class GeoGraph:
             os.makedirs(graph_save_path.parent, exist_ok=True)
             self._save_graph(graph_save_path)
 
+        if self.df.crs is None:
+            self.df.crs = crs
+        else:
+            self.df = self.df.to_crs(crs)
+
         print(
             f"Graph successfully loaded with {self.graph.number_of_nodes()} nodes",
             f"and {self.graph.number_of_edges()} edges.",
@@ -191,6 +197,15 @@ class GeoGraph:
     def df(self) -> gpd.GeoDataFrame:
         """Return dataframe containing polygons."""
         return self._df
+
+    @df.setter
+    def df(self, new_df):
+        self._df = new_df
+
+    @property
+    def crs(self):
+        """Return crs of dataframe."""
+        return self.df.crs
 
     def _load_from_vector_path(
         self,
@@ -388,7 +403,7 @@ class GeoGraph:
             df = df[attributes]
 
         # Reset index to ensure consistent indices
-        df = df.reset_index()
+        df = df.reset_index(drop=True)
         # Using this dict and iterating through it is slightly faster than
         # iterating through df due to the dataframe overhead
         geom: Dict[int, shapely.Polygon] = df.geometry.to_dict()
@@ -424,13 +439,14 @@ class GeoGraph:
                 ]
 
             graph_dict[index] = neighbours
-            # add each polygon as a node to the graph with all attributes
+            # add each polygon as a node to the graph with useful attributes
             self.graph.add_node(
                 index,
                 rep_point=polygon.representative_point().coords[0],
                 area=polygon.area,
                 perimeter=polygon.length,
                 class_label=class_labels[index],
+                bounds=polygon.bounds,
             )
 
         # iterate through the dict and add edges between neighbouring polygons
@@ -455,8 +471,8 @@ class GeoGraph:
             node_list (List[int]): List of integer node indexes in the graph.
             class_label (int): Class label for the resulting node.
             final_index (int, optional): Index to assign to the resulting node.
-            Defaults to None, in which case it becomes the number of nodes in
-            the graph.
+            Defaults to None, in which case it becomes the highest valid index
+            in the dataframe + 1.
 
         Raises:
             ValueError: If there are invalid nodes in `node_list`, or if
@@ -473,36 +489,44 @@ class GeoGraph:
                 "`final_index` cannot be an existing node that is not in `node_list`."
             )
         if final_index is None:
-            final_index = self.graph.number_of_nodes
+            final_index = self.df.last_valid_index()
 
-        adjacency_list = set()
+        # Build set of all neighbours of nodes in node_list, excluding the
+        # nodes in node_list.
+        adjacency_set = set()
         for node in node_list:
-            adjacency_list.update(list(self.graph.neighbors(node)))
-        polygon = unary_union([self._df.at[node, "geometry"] for node in node_list])
+            adjacency_set.update(list(self.graph.neighbors(node)))
+        adjacency_set -= adjacency_set.intersection(node_list)
+        # Build union polygon.
+        polygon = self.df.geometry.iloc[node_list].unary_union
         # Remove nodes from graph and rows from df
         self.graph.remove_nodes_from(node_list)
-        self._df = self._df.drop(node_list)
+        self.df = self.df.drop(node_list)
         # Add final node to graph and df
         self.graph.add_node(
             final_index,
-            rep_point=polygon.representative_point().coords[0],
+            rep_point=polygon.representative_point(),
             area=polygon.area,
             perimeter=polygon.length,
             class_label=class_label,
+            bounds=polygon.bounds,
         )
         data = {"class_label": class_label, "geometry": polygon}
-        missing_cols = {key: None for key in set(self._df.columns) - set(data.keys())}
-        self._df.loc[final_index] = gpd.GeoSeries({**data, **missing_cols})
+        missing_cols = {key: None for key in set(self.df.columns) - set(data.keys())}
+        self.df.loc[final_index] = gpd.GeoSeries(
+            {**data, **missing_cols}, crs=self.df.crs
+        )
+        self.df = self.df.sort_index()
         # Add edges from final node to all neighbours of the original
         self.graph.add_edges_from(
-            zip_longest([final_index], adjacency_list, fillvalue=final_index)
+            zip_longest([final_index], adjacency_set, fillvalue=final_index)
         )
 
     def add_habitat(
         self,
         name: str,
         valid_classes: set,
-        max_travel_distance: float,
+        max_travel_distance: float = 0.0,
     ) -> None:
         """
         Create habitat graph and store it in habitats dictionary.
@@ -519,24 +543,28 @@ class GeoGraph:
             the habitat can travel through non-habitat areas. The habitat graph
             will contain edges between any two nodes that have a class label in
             `valid_classes`, as long as they are less than `max_travel_distance`
-            units apart.
+            units apart. Defaults to 0, which will only create edges between
+            directly neighbouring areas.
         """
         hgraph: nx.Graph = deepcopy(self.graph)
+        hgraph.graph["name"] = name
         # Remove all edges in the graph, then at the end we only have edges
         # between nodes less than `max_travel_distance` apart
         hgraph.clear_edges()
         # Vectorised buffer on the entire df to calculate the expanded polygons
         # used to get intersections.
-        self._df.buff_polygon = self._df.geometry.buffer(max_travel_distance)
+        self.df["buff_polygon"] = self.df.geometry.buffer(max_travel_distance)
         # List of nodes that are not in the habitat
         invalid_indexes: List[int] = []
-        for node in hgraph.nodes:
+        for node in tqdm(
+            hgraph.nodes, desc="Generating habitat graph", total=len(hgraph)
+        ):
             if hgraph.nodes[node]["class_label"] not in valid_classes:
                 invalid_indexes.append(node)
                 continue
             # df.at is a fast df.loc designed only to access single values
-            polygon = self._df.at[node, "geometry"]
-            buff_poly = self._df.at[node, "buff_polygon"]
+            polygon = self.df.at[node, "geometry"]
+            buff_poly = self.df.at[node, "buff_polygon"]
             # Query rtree for all polygons within `max_travel_distance` of the original
             poss_neighbours = list(self.rtree.intersection(buff_poly.bounds))
             for nbr in poss_neighbours:
@@ -544,7 +572,7 @@ class GeoGraph:
                 if hgraph.nodes[nbr]["class_label"] not in valid_classes:
                     continue
                 # Otherwise add the edge with distance attribute
-                nbr_polygon = self._df.at[nbr, "geometry"]
+                nbr_polygon = self.df.at[nbr, "geometry"]
                 if not hgraph.has_edge(node, nbr) and nbr_polygon.intersects(buff_poly):
                     # Since intersects is cheaper than distance, we can save time here
                     if nbr_polygon.intersects(polygon):
@@ -554,7 +582,7 @@ class GeoGraph:
                     hgraph.add_edge(node, nbr, distance=distance)
 
         # Remove expanded polygons from df
-        self._df = self._df.drop(columns=["buff_polygon"])
+        self.df = self.df.drop(columns=["buff_polygon"])
         # Remove non-habitat nodes from habitat graph
         hgraph.remove_nodes_from(invalid_indexes)
         # Add habitat graph to habitats dict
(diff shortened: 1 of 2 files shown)


src/models/geograph.py
```diff
@@ -443,27 +450,143 @@ def merge_nodes(
...
+                    continue
+                # Otherwise add the edge with distance attribute
+                nbr_polygon = self.df.at[nbr, "geometry"]
+                if not hgraph.has_edge(node, nbr) and nbr_polygon.intersects(buff_poly):
+                    # Since intersects is cheaper than distance, we can save time here
+                    if nbr_polygon.intersects(polygon):
+                        distance = 0.0
+                    else:
+                        distance = polygon.distance(nbr_polygon)
+                    hgraph.add_edge(node, nbr, distance=distance)
```
rdnfn, line 546: Great that you're computing distance as well, that will be useful for metrics! We might wan't to make the distance computation optional though, depending on how computationally expensive it is compared to the rest. The idea would be that users might only want to quickly visualise a habitat without metrics. Do you think that would potentially save a lot of time for users?

src/models/geograph.py
```diff
@@ -443,27 +450,143 @@ def merge_nodes(
...
+            zip_longest([final_index], adjacency_set, fillvalue=final_index)
+        )
+
+    def add_habitat(
+        self,
+        name: str,
+        valid_classes: set,
+        max_travel_distance: float = 0.0,
+    ) -> None:
+        """
```
rdnfn, line 495: Super nice function! 🎉  Thanks a lot for implementing it, will be great for a lot of other applications.

This fairly nitpicking but we could consider also accepting `list` besides `set` as the type for the `valid_classes`. My thinking here is that many users (including myself) aren't too familiar with set, so it might add a little confusion for them. It also shouldn't be difficult to do because both because you can use the `set` operator on a set and a list, like `set(set([1,0,1]))`. So just adding the type in the docstring and the line below would do it (as far as I understand it):
```
valid_classes = set(valid_classes)
```

src/models/geograph.py
```diff
@@ -443,27 +450,143 @@ def merge_nodes(
...
+            valid_classes (set): A set of class labels which make up the habitat.
+            max_travel_distance (float): The maximum distance the animal(s) in
+            the habitat can travel through non-habitat areas. The habitat graph
+            will contain edges between any two nodes that have a class label in
+            `valid_classes`, as long as they are less than `max_travel_distance`
+            units apart. Defaults to 0, which will only create edges between
+            directly neighbouring areas.
+        """
+        hgraph: nx.Graph = deepcopy(self.graph)
+        hgraph.graph["name"] = name
```
rdnfn, line 514:
```suggestion
            directly neighbouring areas.
        """
        valid_classes = set(valid_classes)
        
        hgraph: nx.Graph = deepcopy(self.graph)
        hgraph.graph["name"] = name
```

src/models/geograph.py
```diff
@@ -443,27 +450,143 @@ def merge_nodes(
...
+
+        print(
+            f"Habitat successfully loaded with {hgraph.number_of_nodes()} nodes",
+            f"and {hgraph.number_of_edges()} edges.",
         )
+
+    def get_graph_components(
+        self, graph: nx.Graph
+    ) -> Tuple[gpd.GeoDataFrame, List[set]]:
+        """Return a GeoDataFrame with graph components.
```
rdnfn, line 563: Very neat function! Love that it's just four lines  😄

src/models/geograph.py
```diff
@@ -443,27 +450,143 @@ def merge_nodes(
...
+
+        Args:
+            graph (nx.Graph): nx.Graph object. This must be a subgraph of this
+            class's main `graph` attribute.
+
+        Returns:
+            Tuple: A tuple containing the resulting GeoDataFrame and the list of
+            graph components.
+        """
+        components = list(nx.connected_components(graph))
```
rdnfn, line 589: Is `components` a list of lists of node indices?

src/models/geograph.py
```diff
@@ -167,29 +178,45 @@ def graph(self, new_graph):
...
     @property
-    def rtree(self) -> STRtree:
+    def rtree(self) -> rtree_lib.index.Index:
         """Return Rtree object."""
         return self._rtree
 
+    @property
+    def habitats(self) -> Dict[str, nx.Graph]:
+        """Return habitat dictionary."""
+        return self._habitats
```
rdnfn, line 188: Excited for this property 👏

src/models/geograph.py
```diff
@@ -443,27 +450,143 @@ def merge_nodes(
...
+            hgraph.nodes, desc="Generating habitat graph", total=len(hgraph)
+        ):
+            if hgraph.nodes[node]["class_label"] not in valid_classes:
+                invalid_indexes.append(node)
+                continue
+            # df.at is a fast df.loc designed only to access single values
+            polygon = self.df.at[node, "geometry"]
+            buff_poly = self.df.at[node, "buff_polygon"]
+            # Query rtree for all polygons within `max_travel_distance` of the original
+            poss_neighbours = list(self.rtree.intersection(buff_poly.bounds))
```
rdnfn, line 533: You probably checked this, but this obviously might have quite a few non-relevant neighbours. Have you considered creating an rtree for the habitat? (Just wondering, don't mind it either way - I am not familiar with the complexity of creating an rtree)

src/models/geograph.py
rdnfn: One more thing: would it also be possible to save the inputs for the habitat? It might be handy to have both the max_travel_distance and valid_classes somewhere. (I am thinking to create a visualisation later). Maybe the habitat property could be a list of dict of dicts 
```
{name:{graph:nx.graph,max_travel_distance:int,valid_classes:set},...}
```
It's easier to change this now than later ... but maybe you have a better idea of how to store this info?

src/models/geograph.py
```diff
@@ -443,27 +450,143 @@ def merge_nodes(
...
+
+        Creates a habitat subgraph of the main GeoGraph that only contains edges
+        between nodes in `valid_classes` as long as they are less than
+        `max_travel_distance` apart. All nodes which are not in `valid_classes`
+        are not in the resulting habitat graph.
+
+        Args:
+            name (str): The name of the habitat.
+            valid_classes (set): A set of class labels which make up the habitat.
+            max_travel_distance (float): The maximum distance the animal(s) in
```
rdnfn, line 506: Would there be a simple way to take this argument in meters? What the unit at the moment?

Generally, it doesn't need to be perfect. Like if there is slight variations with the coordinates across a graph of like less than a few meters I think it should be fine.

src/models/geograph.py
```diff
@@ -86,13 +83,13 @@ def __init__(
             not save the polygonised data.
             tolerance (float, optional): Adds edges between neighbours that are
             at most `tolerance` units apart. Defaults to 0.
+            crs (str, optional): Coordinate reference system to set on the
+            resulting dataframe. Defaults to None.
```
rdnfn, line 87: Super nit: you probably changed this default later... it defaults to PREFERRED_CRS which is UTM35N (?) at the moment?

rdnfn (approved):
Excellent stuff, especially thanks for implementing those habitat and component methods! 🌳 🚀 

I just added some very minor points, most aimed at improving user experience slightly, but all very optional. Happy to merge it as is though.

src/models/geograph.py
```diff
@@ -321,51 +348,36 @@ def _load_from_dataframe(
...
             raise ValueError("`geometry` must be a column in the dataframe.")
-        # If no attribute list is given, all
-        # columns in df are used
-        if attributes is None:
-            attributes = df.columns.tolist()
-        elif not set(attributes).issubset(df.columns):
-            raise ValueError("`attributes` must only contain column names in `df`.")
 
+        # Reset index to ensure consistent indices
+        df = df.reset_index(drop=True)
```
Croydon-Brixton, line 370: By default `.reset_index` creates a copy of the original dataframe right? (So we're not running at risk of modifying variables outside of the function scope)

src/models/geograph.py
```diff
@@ -321,51 +348,36 @@ def _load_from_dataframe(
...
-            id(polygon): index for index, polygon in geom.items()
-        }
+        class_labels: List[int] = df.class_label.tolist()
         # Build Rtree from geometry
-        tree = STRtree(geom.values())
-        # this dict maps polygon indices in df to a list
-        # of neighbouring polygon indices
+        tree = rtree_lib.index.Index()
+        for index, polygon in geom.items():
+            tree.insert(index, polygon.bounds)
```
Croydon-Brixton, line 378: Out of curiosity, this part is super fast right? (From the experience you had so far with respect to the other steps involved in graph creation)
(because when we reproject we probably have to do it as well, and when we merge graphs/add nodes we will also have to update the `rtree`)

Croydon-Brixton: On a related note, what do you think of factoring out this part of the method as `def _build_rtree(self ...)` for easier readability and in case we want to rebuild `rtrees` in other functions in future?

src/models/geograph.py
```diff
@@ -379,31 +391,26 @@ def _load_from_dataframe(
                 new_polygon = polygon.buffer(tolerance)
                 # find the indexes of all polygons which intersect with this one
                 neighbours: List[int] = [
```
Croydon-Brixton, line 393: Related to the above:
Should we factor out this method as `def _find_nearest_neighbors(self, ...)` to make the load function slightly easier to read in case we want to use it elsewhere?

src/models/geograph.py
```diff
@@ -443,27 +450,143 @@ def merge_nodes(
...
-            [self.graph.nodes[node]["geometry"] for node in node_list]
-        )
-
+            adjacency_set.update(list(self.graph.neighbors(node)))
+        adjacency_set -= adjacency_set.intersection(node_list)
+        # Build union polygon.
+        polygon = self.df.geometry.iloc[node_list].unary_union
+        # Remove nodes from graph and rows from df
         self.graph.remove_nodes_from(node_list)
+        self.df = self.df.drop(node_list)
```
Croydon-Brixton, line 468: Will this lead to potentially missing indices?
(I'm thinking e.g. `1, 2, 3, 5, 6, 8, ...` ? If yes, this could be a problme with how we check `final_index` above)

src/models/geograph.py
```diff
@@ -443,27 +450,143 @@ def merge_nodes(
             and self.graph.has_node(final_index)
         ):
             raise ValueError(
-                "`final_index` must not be an existing node not in `node_list`."
+                "`final_index` cannot be an existing node that is not in `node_list`."
             )
         if final_index is None:
```
Croydon-Brixton, line 455: Potential simplification together with the above:

```suggestion
if final_index is None:
    final_index = self.df.last_valid_index()
elif final_index not in node_list and self.graph.has_node(final_index):
    raise ValueError(....)
```

src/models/geograph.py
```diff
@@ -443,27 +450,143 @@ def merge_nodes(
...
+        data = {"class_label": class_label, "geometry": polygon}
+        missing_cols = {key: None for key in set(self.df.columns) - set(data.keys())}
+        self.df.loc[final_index] = gpd.GeoSeries(
+            {**data, **missing_cols}, crs=self.df.crs
         )
+        self.df = self.df.sort_index()
+        # Add edges from final node to all neighbours of the original
         self.graph.add_edges_from(
-            zip_longest([final_index], adjacency_list, fillvalue=final_index)
+            zip_longest([final_index], adjacency_set, fillvalue=final_index)
```
Croydon-Brixton, line 486: Maybe I missed it, but we'll also need to update the `rtree` right?

Croydon-Brixton (commented):
Nice one, sorry for taking longer to review - I mainly added a few questions regarding `rtree` updates and copy vs. reference questions 😄 👍

src/models/geograph.py
```diff
@@ -321,51 +348,36 @@ def _load_from_dataframe(
...
         """
         if tolerance < 0.0:
             raise ValueError("`tolerance` must be greater than 0.")
-        if (
-            attributes is not None
-            and "class_label" not in attributes
-            or "class_label" not in df.columns
-        ):
+        if "class_label" not in df.columns:
             raise ValueError("`class_label` must be a column in the dataframe.")
```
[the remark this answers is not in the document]
herbiebradley: Good idea - I think the cleanest option may be to pass a dict of columns to rename.

src/models/geograph.py
herbiebradley: It is a list of sets of node indices.

src/models/geograph.py
```diff
@@ -167,29 +178,45 @@ def graph(self, new_graph):
...
+        return self._habitats
+
+    @property
+    def df(self) -> gpd.GeoDataFrame:
+        """Return dataframe containing polygons."""
+        return self._df
+
+    @df.setter
+    def df(self, new_df):
+        self._df = new_df
```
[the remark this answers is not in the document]
Croydon-Brixton: Ok, actually found my answer here: 
https://stackoverflow.com/a/36943813

TL;DR: 
It's mainly useful when changing the underlying functionality of how variables are fetched (e.g. by checking against valid options first or so)

src/models/geograph.py
Croydon-Brixton: @herbiebradley 
I read up a bit more on this and `geopandas` actually supports a native `rtree` (either in `geos` (written in C) via `pygeos` or with the same `rtree` python library we're using, depending on whether geos is installed).

--> This could be a simple fix to updating and creating the `rtree` and also to transforming it under `crs` operations?
What do you think?


Here's the [gpd code I refer to](https://github.com/geopandas/geopandas/blob/57262a8821ab3c0f348279b6d6fb14512dde8982/geopandas/sindex.py)

src/models/geograph.py
herbiebradley: The fastest way to creating an rtree for the habitat would be to deepcopy the main one and remove all non-habitat nodes. This would take some upfront computation time, so for this to be faster eventually would require both quite a small habitat and for us to frequently use some other function for habitats which iterates through the results of rtree queries. I'm not sure if there will be another function like that, so I think we'll leave it for now.

src/models/geograph.py
herbiebradley: The distances are unitless at the moment to allow for passing in data which is not in metres, but I think all the distances in our data are metres anyway so this argument should already be in metres.

src/models/geograph.py
Croydon-Brixton: Doing some tests on this now and it works pretty well, we would simply need to replace/bind `.rtree` with `df.sindex` and use the `.sindex` interface (which also provides a `query_bulk` option which is vectorized if we use geos under the hood).

image


It also handles dropping nodes automatically 

image

src/models/geograph.py
```diff
@@ -321,51 +348,36 @@ def _load_from_dataframe(
...
-        if attributes is None:
-            attributes = df.columns.tolist()
-        elif not set(attributes).issubset(df.columns):
-            raise ValueError("`attributes` must only contain column names in `df`.")
 
+        # Reset index to ensure consistent indices
+        df = df.reset_index(drop=True)
+        # Using this dict and iterating through it is slightly faster than
+        # iterating through df due to the dataframe overhead
         geom: Dict[int, shapely.Polygon] = df.geometry.to_dict()
```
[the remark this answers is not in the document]
herbiebradley: 1. Yes, it should just copy references (a shallow copy). I think the answer to your second question is yes, but I'm not sure why this would be a concern since this `geom` dict is not used after we build the graph.
2. Yes, that's correct. Pandas has very fast methods of accessing multiple rows, but the reason I'm using this `geom` object here is because I tested just looping over the rows of the dataframe and it was ~5-10% slower - you really want to try and stick to vectorised ops with dataframes.

Thanks for bringing this to my attention, since I just realised that now we're doing `reset_index` I can change the dict to a list, which is slightly faster for index lookups and uses less memory than a dict.

src/models/geograph.py
herbiebradley: Thanks for reminding me - I forgot about that!

src/models/geograph.py
herbiebradley: I tested it for some large habitats and it does shave off a decent amount of time, so I'll make it optional. If it is a very large habitat though it will still take ~20-40 seconds to load, but small habitats with only a few thousand nodes are usually less than 10s.

src/models/geograph.py
herbiebradley: Yes, this is very fast - and factoring it out seems like a good idea.

src/models/geograph.py
herbiebradley: OK, we decided on:
1. Converting to the passed CRS at the beginning of the `_load_from_dataframe` function
2. Adding a warning to the docstring.

src/models/geograph.py
```diff
@@ -42,17 +42,17 @@ class GeoGraph:
     def __init__(
         self,
         data,
-        attributes: Optional[List[str]] = None,
         graph_save_path: Optional[Union[str, os.PathLike]] = None,
         raster_save_path: Optional[Union[str, os.PathLike]] = None,
         tolerance: float = 0.0,
+        crs: str = PREFERRED_CRS,
```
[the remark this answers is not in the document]
herbiebradley: You mean make it a mandatory argument? That sounds fine.

src/models/geograph.py
```diff
@@ -152,6 +158,11 @@ def __init__(
             os.makedirs(graph_save_path.parent, exist_ok=True)
             self._save_graph(graph_save_path)
 
+        if self.df.crs is None:
+            self.df.crs = crs
+        else:
+            self.df = self.df.to_crs(crs)
```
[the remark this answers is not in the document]
herbiebradley: From our discussion: the bounds don't change when we reproject, but this is not an issue if we only project once (before we build the `rtree`). However, some things are simplified if we switch to GeoPandas built-in rtree (`sindex` I think) so I'll look into that.

src/models/geograph.py
herbiebradley: I'm changing this to remove the default so it's a mandatory argument.

src/models/geograph.py
rdnfn: Great, thanks for testing it!

src/models/geograph.py
herbiebradley: Yeah, all dataframe functions usually have an `inplace` option which is `False` by default but is generally a bad idea to set to `True` https://stackoverflow.com/questions/45570984/in-pandas-is-inplace-true-considered-harmful-or-not

src/models/geograph.py
herbiebradley: Good catch!

src/models/geograph.py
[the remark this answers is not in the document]
herbiebradley: Since we don't really have any need to do that, I've changed the `graph` and `df` properties to be attributes and the rest can remain properties because we want them to be read-only.

src/models/geograph.py
herbiebradley: I refactored the `.buffer` outside the for loop to make it vectorised so this is now difficult to do unless we ignore the `tolerance` argument. With the new spatial index rtree the code is much more compact anyway.

src/models/geograph.py
herbiebradley: Update: in 8c0a035 I changed to the spatial index rtree, so factoring out the build_rtree is now unnecessary since the tree gets built whenever you make a `GeoDataFrame`.

herbiebradley:
Major commit: in 8c0a035 I have switched over to the GeoPandas spatial index rtree, which is now bound to `self.rtree` so it behaves just as before but with the extra methods `query and `bulk_query`. Loading speed improved by ~80% and the `add_habitats` function is also noticeably faster. I also implemented a dataclass for each habitat, so the habitats main dict looks like this:
habitats

And you access each habitat just by calling the dict with the name, like `geograph.habitats["Wolf"]`. I also resolved the minor things mentioned above.

src/models/geograph.py
```diff
@@ -223,17 +249,17 @@ def _load_from_raster_path(
             but Shapefiles also work.
 
         Returns:
-            STRtree: The Rtree object used to build the graph.
+            rtree_lib.index.Index: The Rtree object used to build the graph.
         """
         with rasterio.Env():
```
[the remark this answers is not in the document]
herbiebradley: Thanks, I removed it.

src/models/geograph.py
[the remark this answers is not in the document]
herbiebradley: In 8c0a035 I switched to the GeoPandas built-in rtree so the bounds should change when we reproject.

src/models/geograph.py
herbiebradley: Done!

src/models/geograph.py
herbiebradley: Done!

src/models/geograph.py
```diff
@@ -36,16 +35,26 @@
...
+    add_distance: bool
+
+
 class GeoGraph:
     """Class for the fragmentation graph."""
 
     def __init__(
         self,
         data,
+        crs: str,
```
Croydon-Brixton, line 53: Allow passing of custom built CRS objects in addition to epsg strings.

```suggestion
        crs: Union[str, pyproj.CRS], 
```

src/models/geograph.py
```diff
@@ -36,16 +35,26 @@
 )
 
 
+@dataclass
```
Croydon-Brixton, line 38: Super cool!

src/models/geograph.py
```diff
@@ -97,15 +112,17 @@ def __init__(
             polygonisation creating invalid geometries.
         """
         super().__init__()
-        self._graph = nx.Graph()
-        self._habitats: Dict[str, nx.Graph] = {}
+        self.graph = nx.Graph()
+        self._habitats: Dict[str, Habitat] = {}
+        self._crs: str = crs
```
Croydon-Brixton, line 117:
```suggestion
        self._crs: Union[str, pyproj.CRS] = crs
```

Croydon-Brixton (approved):
Looks great to me 💯 👍  Thanks for going the extra mile! 😍 

img

PR #32, opened 2021-02-27
Feature/graph analysis perf

herbiebradley:
I first made this branch to test out PyGeos and some other performance things, but it turns out PyGeos is really slow. This is partly because it uses a different rtree in the `sindex`, but even the things like `buffer` on a GeoDataFrame which are supposed to be where it's useful were significantly slower than the normal version.

So this is just a quick PR with prepared geometries added to the `add_habitat` function - this speeds up the function by 50-70% by making the main loop practically instant. Now basically all of the computation time is taken up by `buffer` at the beginning, so if you set `max_travel_distance=0` the whole thing is very fast.


src/models/geograph.py
```diff
@@ -518,22 +520,24 @@ def add_habitat(
                 invalid_indexes.append(node)
                 continue
 
-            polygon = polygons[node]
+            polygon = prep(polygons[node])
```
Croydon-Brixton, line 523: I think we can move this into the else branch of your if-else below (line525) directly right? (not that it makes a big difference ^^)

src/models/geograph.py
```diff
@@ -518,22 +520,24 @@ def add_habitat(
...
             if max_travel_distance > 0:
-                buff_poly = buff_polygons[node]
+                buff_poly_bounds = buff_polygons[node].bounds
+                buff_poly = prep(buff_polygons[node])
```
Croydon-Brixton, line 526: Sweet, I like the prep - just looked up what it does under the hood, essentially a combo of smart caching & short circuit eval to speed up e.g. intersection checks - great idea to use it!

For reference:
- https://sgillies.net/2009/01/26/efficient-batch-operations-for-shapely.html
- http://lin-ear-th-inking.blogspot.com/2007/08/preparedgeometry-efficient-batch.html

Speed ups are expected:
- when multiple intersects are performed 
- particularly as polys get more complex.
Speed ups of 4x to 40x have apparently been observed in various cases

src/models/geograph.py
```diff
@@ -518,22 +520,24 @@ def add_habitat(
...
             else:
                 buff_poly = polygon
+                buff_poly_bounds = polygons[node].bounds
             # Query rtree for all polygons within `max_travel_distance` of the original
-            for nbr in self.rtree.intersection(buff_poly.bounds):
+            for nbr in self.rtree.intersection(buff_poly_bounds):
                 # If a node is not a habitat class node, don't add the edge
                 if hgraph.nodes[nbr]["class_label"] not in valid_classes_set:
```
Croydon-Brixton, line 529: Could we drop the invalid classes as a bulk operation in hgraph at the outset of the function? That might give us another slight speed up (eg from avoiding the check and python-looping over unnecessary geoms here)

Croydon-Brixton (approved):
Awesome, great job thanks a lot @herbiebradley 🎉🎉

Croydon-Brixton:
Looks good to merge to me

src/models/geograph.py
herbiebradley: Good point - I've added this and it makes the tqdm iteration counter more accurate but makes absolutely no difference to speed as far as I can tell. Turns out those checks are really fast and you have to do some upfront computation to get the indices of invalid nodes.

PR #33, opened 2021-02-28
Feature/fragmentation baseline

### LINKED ISSUE #9: Implementation of a baseline
EDIT:
Remaining things to be done:

- [x] Get ESA CCI data as test data
- [x] Define region of interest
- [x] Clip raster to region of interest
- [x] Separate zones
- [x] Reproject to suitable CRS
- [x] Perform pylandstats analysis on timeseries from 1992 to 2015
- [x] Sanity checks
- [ ] Transfer final pipeline code from jupyter notebooks to python script
- [ ] Finalize plotting procedures and add to python script

Croydon-Brixton:
This PR closes #9.

Features:
- Time-series calculation of conventional fragmentation statistics in user defined zones 
- Ouptut is transformed into multidimensional array for plotting
- Short demo notebook `notebooks/demo-1-conventional-fragmentation-analysis.ipynb` shows how to perform conventional fragmentation analysis and plot result.


src/models/fragmentation_baseline.py
```diff
@@ -0,0 +1,249 @@
...
+        fragmentation_statistics=fragmentation_statistics,
+        landcover_classes=landcover_classes,
+        rois=rois,
+        rois_index_col=rois_index_col,
+    )
+
+    # Save data to disk for further analysis
+    if save_path is not None:
+        fragmentation_data.to_netcdf(save_path)
+        save_path.chmod(0o664)
```
herbiebradley, line 246: This makes the file read-only for others, right? Maybe we should add this to the save code in `geograph.py`?

herbiebradley (approved):
Looks good! I noticed in the notebook that there are a lot of nan values in the resulting array, do you know why this happens?

rdnfn (approved):
Looks good to me too! 🚀 These functions will be great for a direct comparison between pylandstats and our very own graph methods.

src/models/fragmentation_baseline.py
Croydon-Brixton: This changes the file permissions to `-rw-rw-r--` (owner read & write, group read & write, others read)
(By default stuff on the gws is created as `-rw-r--r--` (644).

I like [this site](https://chmodcommand.com/chmod-664/) btw - super useful for crafting chmod commands 😉

Croydon-Brixton: Yes, controlling permissions for geograph could be useful - do you know if ppl need to have `execute` right to load a pickle file? (then we might want `6644` or `6664`)

Croydon-Brixton:
> Looks good! I noticed in the notebook that there are a lot of nan values in the resulting array, do you know why this happens?

Yes, that's actually because lots of the ESA CCI classes do not occur in the Chernobyl data at all (I always insert a `nan` whenever a class doesn't occur. Some classes do not exist at first and suddenly pop up after a certain year (for example the urban class only exists post 2000 interestingly ^^). In this case for urban for example the first half of the array are again `nan` and the second half have actual values.

Croydon-Brixton:
> Looks good to me too! 🚀 These functions will be great for a direct comparison between pylandstats and our very own graph methods.
> 
> EDIT: just one very minor optional point: when publishing the notebook (notebooks/demo-1-conventional-fragmentation-analysis.ipynb) for others to read it might be useful to remove the output with all the nan warnings, because at the moment it's a long scroll down without much information. Maybe changing the log level for warnings, and adding a comment in the code.

Good point - I'll clean that up in the NB

src/models/fragmentation_baseline.py
herbiebradley: I believe you don't need `execute` to load pickle files. But do you know if there is any practical difference on JASMIN for 664 vs 644 (i.e. are all users in a gws part of the same group)?

herbiebradley (approved):
Looks good, I like the notebook organisation ;)

src/models/fragmentation_baseline.py
Croydon-Brixton: Yes, I think all users of the `gws` are in the same group as far as I understand. I couldn't rename some 644 files for example, but could rename 664 ones.

Croydon-Brixton:
@herbiebradley @rdnfn : Only changes: cleaned up the notebook as suggested by @rdnfn  and `isort`ed the imports in `fragmentation-baseline`.

herbiebradley (approved):
👍

PR #40, opened 2021-03-07
Implementing unet model

Crystallee1104:
The train.py and the files in unet_ folder are those for implementing unet model. 

Currnetly, there are several problems for me to process data for unet and seek help:

1. When do BasicDataset function(in src.unet_.dataset.py), image shpae become 5 dimension (2, 19, 681, 1086, 12) (batch size, yr, y, x, band). But the dimension in the model is 4 dimension (64, 3, 3, 3);

2. The input size in UNet model should be (N, Cin, H, W): N is a batch size, CC denotes a number of channels, 
H is a height of input planes in pixels, and W is width in pixels (https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html#torch.nn.Conv2d). 
But input x and y are (yr, y, x, band), and the order should be changed.  Maybe permute function in torch.Tensor can help (https://pytorch.org/docs/stable/tensors.html#torch.Tensor.permute), but not sure how to use it;

3. Plan to use 3 channels in UNet model, represented RGB images. But the x_tr seems to have 12 bands (3 RGB bands * 4 seasons), and will it conflict with 3 bands in UNet (?);

4. Still working on clarify whether n_classes requal to the number of classified land cover classes that we want.

### DIFF
--- a/train.py
+++ b/train.py
@@ -84,14 +84,14 @@ The current problems for implementing UNet are:
 1. When do BasicDataset function(in src.unet_.dataset.py), image shpae become 5 dimension (2, 19, 681, 1086, 12) (batch size, yr, y, x, band).
 But the dimension in the model is 4 dimension (64, 3, 3, 3);
 
-2. The input size in UNet model should be (N, Cin, H, W): N is N is a batch size, CC denotes a number of channels, 
+2. The input size in UNet model should be (N, Cin, H, W): N is a batch size, CC denotes a number of channels, 
 H is a height of input planes in pixels, and W is width in pixels (https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html#torch.nn.Conv2d). 
 But input x and y are (yr, y, x, band), and the order should be changed. 
 Maybe permute function in torch.Tensor can help (https://pytorch.org/docs/stable/tensors.html#torch.Tensor.permute), but not sure how to use it;
 
-3. Plan to use 3 channels in UNet model, represented RGB images. But the x_tr seems to have 12 bands and be reduced to 3 RGB bands.
+3. Plan to use 3 channels in UNet model, represented RGB images. But the x_tr seems to have 12 bands (3 RGB bands * 4 seasons), and will it conflict with 3 bands in UNet (?);
 
-4. Still working on clarify whether n_classes requal to the number of classified land cover classes that we want
+4. Still working on clarify whether n_classes requal to the number of classified land cover classes that we want.
 '''
 
 if __name__ == '__main__':
--- a/train.py
+++ b/train.py
@@ -20,7 +20,7 @@ from src.constants import ESA_LANDCOVER_DIR, GWS_DATA_DIR, SAT_DIR
 from src.preprocessing.esa_compress import compress_esa
 from src.preprocessing.load_landsat_esa import return_xy_np_grid, return_x_y_da
 from src.unet_.dataset import BasicDataset
-
+#x represents landsat data, and y represents ESA CCI LC.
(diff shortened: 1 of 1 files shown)


train.py
```diff
@@ -84,14 +84,14 @@ def train_net(net,
 1. When do BasicDataset function(in src.unet_.dataset.py), image shpae become 5 dimension (2, 19, 681, 1086, 12) (batch size, yr, y, x, band).
 But the dimension in the model is 4 dimension (64, 3, 3, 3);
 
-2. The input size in UNet model should be (N, Cin, H, W): N is N is a batch size, CC denotes a number of channels, 
+2. The input size in UNet model should be (N, Cin, H, W): N is a batch size, CC denotes a number of channels,
```
Croydon-Brixton, line 87: For the training and preduction I suggest splitting up the full image into tiles (e.g. of size 128 x 128 or 256x 256 or so) - In the ArcGIS demo they use 400 x 400 
https://developers.arcgis.com/python/sample-notebooks/land-cover-classification-using-unet/

train.py
```diff
@@ -84,14 +84,14 @@ def train_net(net,
 1. When do BasicDataset function(in src.unet_.dataset.py), image shpae become 5 dimension (2, 19, 681, 1086, 12) (batch size, yr, y, x, band).
 But the dimension in the model is 4 dimension (64, 3, 3, 3);
```
Croydon-Brixton, line 85: Does the Unet model you're using have an option to change the number of channels and intput dim?

train.py
```diff
@@ -84,14 +84,14 @@ def train_net(net,
 1. When do BasicDataset function(in src.unet_.dataset.py), image shpae become 5 dimension (2, 19, 681, 1086, 12) (batch size, yr, y, x, band).
```
Croydon-Brixton, line 84: I guess we probably want to get rid of the year dimension 😉 but haven't looked at the `dataset.py` file in detail yet

Crystallee1104: Got it.
And the processed data is based on @sdat2 's works on preprocessing data for xgboots. It will be clearer about the data if also have a look on unet.py in src.models and src.preprocessing folder.

train.py
```diff
@@ -84,14 +84,14 @@ def train_net(net,
...
 But the dimension in the model is 4 dimension (64, 3, 3, 3);
 
-2. The input size in UNet model should be (N, Cin, H, W): N is N is a batch size, CC denotes a number of channels, 
+2. The input size in UNet model should be (N, Cin, H, W): N is a batch size, CC denotes a number of channels, 
 H is a height of input planes in pixels, and W is width in pixels (https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html#torch.nn.Conv2d). 
 But input x and y are (yr, y, x, band), and the order should be changed. 
 Maybe permute function in torch.Tensor can help (https://pytorch.org/docs/stable/tensors.html#torch.Tensor.permute), but not sure how to use it;
 
-3. Plan to use 3 channels in UNet model, represented RGB images. But the x_tr seems to have 12 bands and be reduced to 3 RGB bands.
+3. Plan to use 3 channels in UNet model, represented RGB images. But the x_tr seems to have 12 bands (3 RGB bands * 4 seasons), and will it conflict with 3 bands in UNet (?);
```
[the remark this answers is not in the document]
Crystallee1104: I'll change the n_channels in Unet to 12.

train.py
Crystallee1104: The Unet model can change the number of channels and the number of classes, but I don't think it can change the input dim

PR #42, opened 2021-03-09
Feature/graph analysis metrics

herbiebradley:
This PR adds metrics to GeoGraph by defining a Metric dataclass. A Metric object is then returned when a metric is calculated; each metric calculation function accepts only a GeoGraph, so that you can calculate any metric by calling the `get_metric` method on any GeoGraph and passing only the name of the metric. The PR also changes Habitats and Components to be sub-classes of GeoGraph, so you can now have recursively nested HabitatGeoGraphs, each with a ComponentGeoGraph, and you can calculate metrics on all of them!

Other features:

- A `merge_classes` function which can be used to avoid some extra lines when combining class labels in the dataframe.
- Tested support for string class labels as well as integer ones.
- An `apply_to_habitats` function which allows you to apply a GeoGraph method, like `merge_nodes`, to every HabitatGeoGraph attached to a GeoGraph in one go, assuming each application uses the same arguments.
- Further speedup in `add_habitat`.
- Improved `get_graph_components` function, which now supports optionally calculating the union polygons for each component (by default this is not done for a GeoGraph, but is done for a HabitatGeoGraph since habitats are often small enough that this is fairly cheap).
- Habitat-specific saving and loading.
- Implemented \_\_eq\_\_ for GeoGraphs which checks if the two GeoGraphs are isomorphic.
- Similarly, you can compare Metric objects with < <= == >= > and they will work as you would expect by comparing the underlying values.
- ComponentGeoGraphs support optionally adding an edge between every pair of nodes (each node represents a component in the original graph) with the distance between the node polygons as an edge attribute. This is very slow for any habitat with more than about 100 components, so is disabled by default, but it is necessary for the `avg_component_isolation` metric.

### DIFF
--- a/src/models/geograph.py
+++ b/src/models/geograph.py
@@ -7,24 +7,28 @@ from __future__ import annotations
 
 import bz2
 import gzip
+import inspect
 import os
 import pathlib
 import pickle
 from copy import deepcopy
-from dataclasses import dataclass
 from itertools import zip_longest
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Union
 
 import geopandas as gpd
 import networkx as nx
 import numpy as np
+import pandas as pd
 import pyproj
 import rasterio
 import shapely
 from shapely.prepared import prep
 from src.data_loading import rasterio_utils
-from src.models import binary_graph_operations
+from src.models import binary_graph_operations, metrics
+from src.models.metrics import Metric
 from tqdm import tqdm
+
+pd.options.mode.chained_assignment = None  # default='warn'
 
 VALID_EXTENSIONS = (
     ".pickle",
@@ -38,15 +42,6 @@ VALID_EXTENSIONS = (
     ".geotif",
     ".geotiff",
 )
-
-
-@dataclass
-class Habitat:
-    name: str
-    graph: nx.Graph
-    valid_classes: List[int]
-    max_travel_distance: float
-    add_distance: bool
 
 
 class GeoGraph:
@@ -118,10 +113,11 @@ class GeoGraph:
         """
         super().__init__()
         self.graph = nx.Graph()
-        self._habitats: Dict[str, Habitat] = {}
+        self.habitats: Dict[str, HabitatGeoGraph] = {}
         self._crs: Optional[Union[str, pyproj.CRS]] = crs
         self._columns_to_rename: Optional[Dict[str, str]] = columns_to_rename
         self._tolerance: float = tolerance
+        self.metrics: Dict[str, Metric] = {}
 
         if raster_save_path is not None:
             raster_save_path = pathlib.Path(raster_save_path)
@@ -168,28 +164,25 @@ class GeoGraph:
         # Save resulting graph, if we didn't load from graph.
         if not load_from_graph and graph_save_path is not None:
             graph_save_path = pathlib.Path(graph_save_path)
-            if graph_save_path.suffix not in (".pickle", ".pkl", ".gz", ".bz2"):
-                raise ValueError(
-                    """Argument `graph_save_path` should be a pickle file or
-                    compressed file."""
-                )
             os.makedirs(graph_save_path.parent, exist_ok=True)
-            self._save_graph(graph_save_path)
+            self.save_graph(graph_save_path)
+
+        self.components = self.get_graph_components(calc_polygons=False)
 
         print(
             f"Graph successfully loaded with {self.graph.number_of_nodes()} nodes",
             f"and {self.graph.number_of_edges()} edges.",
         )
 
+    def __eq__(self, o: object) -> bool:
+        if not isinstance(o, GeoGraph):
+            return False
+        return nx.is_isomorphic(self.graph, o.graph)
+
     @property
     def rtree(self):
         """Return Rtree object."""
         return self.df.sindex
-
-    @property
-    def habitats(self) -> Dict[str, Habitat]:
-        """Return habitat dictionary."""
-        return self._habitats
 
     @property
     def crs(self):
@@ -202,7 +195,7 @@ class GeoGraph:
 
         Note: Uses `iloc` type indexing.
         """
-        return self.df.class_label.values
+        return self.df["class_label"].values
 
     @property
     def geometry(self):
@@ -210,7 +203,7 @@ class GeoGraph:
 
         Note: Uses `iloc` type indexing.
         """
-        return self.df.geometry.values
+        return self.df["geometry"].values
 
     def _load_from_vector_path(
         self,
@@ -321,14 +314,22 @@ class GeoGraph:
         self.graph = data["graph"]
         return data["dataframe"]
 
-    def _save_graph(self, save_path: pathlib.Path) -> None:
+    def save_graph(self, save_path: pathlib.Path) -> None:
         """
         Save graph with attributes and dataframe as pickle file. Can be compressed.
 
         Args:
             save_path (pathlib.Path): Path to a pickle file. Can be compressed
             with gzip or bz2 by passing filenames ending in `gz` or `bz2`.
+
+        Raises:
+            ValueError: If `save_path` is not a pickle, gz, or bz2 file.
         """
+        if save_path.suffix not in (".pickle", ".pkl", ".gz", ".bz2"):
+            raise ValueError(
+                """Argument `save_path` should be a pickle file or
+                compressed file."""
+            )
         data = {"graph": self.graph, "dataframe": self.df}
         if save_path.suffix == ".bz2":
             with bz2.BZ2File(save_path, "wb") as bz2_file:
@@ -383,15 +384,16 @@ class GeoGraph:
         df = df.reset_index(drop=True)
         # Using this list and iterating through it is slightly faster than
         # iterating through df due to the dataframe overhead
-        geom: List[shapely.Polygon] = df.geometry.tolist()
-        class_labels: List[int] = df.class_label.tolist()
+        geom: List[shapely.Polygon] = df["geometry"].tolist()
         # this dict maps polygon row numbers in df to a list
         # of neighbouring polygon row numbers
         graph_dict = {}
 
         if tolerance > 0:
             # Expand the borders of the polygons by `tolerance```
-            new_polygons: List[shapely.Polygon] = df.geometry.buffer(tolerance).tolist()
+            new_polygons: List[shapely.Polygon] = (
+                df["geometry"].buffer(tolerance).tolist()
+            )
         # Creating nodes (=vertices) and finding neighbors
         for index, polygon in tqdm(
             enumerate(geom),
@@ -413,7 +415,6 @@ class GeoGraph:
                 rep_point=polygon.representative_point(),
                 area=polygon.area,
                 perimeter=polygon.length,
-                class_label=class_labels[index],
                 bounds=polygon.bounds,
             )
         # iterate through the dict and add edges between neighbouring polygons
@@ -429,7 +430,10 @@ class GeoGraph:
         return df
 
     def merge_nodes(
-        self, node_list: List[int], class_label: int, final_index: int = None
+        self,
+        node_list: List[int],
+        class_label: Union[int, str],
+        final_index: int = None,
     ) -> None:
         """
         Merge a list of nodes in the graph together into a single node.
@@ -439,19 +443,20 @@ class GeoGraph:
 
         Args:
             node_list (List[int]): List of integer node indexes in the graph.
-            class_label (int): Class label for the resulting node.
+            class_label (int or str): Class label for the resulting node.
             final_index (int, optional): Index to assign to the resulting node.
             Defaults to None, in which case it becomes the highest valid index
             in the dataframe + 1.
 
         Raises:
-            ValueError: If there are invalid nodes in `node_list`, or if
-            `final_index` is an existing node not in `node_list`.
+            ValueError: If `final_index` is an existing node not in `node_list`,
+            or if `node_list` does not contain any existing nodes.
         """
-        if not all(self.graph.has_node(node) for node in node_list):
-            raise ValueError("`node_list` must only contain valid nodes.")
+        node_list = [node for node in node_list if self.graph.has_node(node)]
+        if len(node_list) == 0:
+            raise ValueError("`node_list` must contain at least one node in the graph.")
         if final_index is None:
-            final_index = self.df.last_valid_index()
+            final_index = self.df.last_valid_index() + 1
         elif final_index not in node_list and self.graph.has_node(final_index):
             raise ValueError(
                 "`final_index` cannot be an existing node that is not in `node_list`."
@@ -464,7 +469,7 @@ class GeoGraph:
             adjacency_set.update(list(self.graph.neighbors(node)))
         adjacency_set -= adjacency_set.intersection(node_list)
         # Build union polygon.
-        polygon = self.df.geometry.loc[node_list].unary_union
+        polygon = self.df["geometry"].loc[node_list].unary_union
         # Remove nodes from graph and rows from df
         self.graph.remove_nodes_from(node_list)
         self.df = self.df.drop(node_list)
@@ -474,7 +479,6 @@ class GeoGraph:
             rep_point=polygon.representative_point(),
             area=polygon.area,
             perimeter=polygon.length,
-            class_label=class_label,
             bounds=polygon.bounds,
         )
         data = {"class_label": class_label, "geometry": polygon}
@@ -488,25 +492,48 @@ class GeoGraph:
             zip_longest([final_index], adjacency_set, fillvalue=final_index)
         )
 
+    def merge_classes(
+        self, class_list: List[Union[str, int]], new_name: Union[str, int]
+    ) -> None:
+        """
+        Merge multiple classes together into one by renaming the class labels.
+
+        Args:
+            new_name (Union[str, int]): The new name for the combined class,
+            either a string or an int.
+            class_list (List): The list of names of class labels to combine.
+            Every name in the list must be in the GeoGraph.
+
+        Raises:
+            ValueError: If `class_list` contains a class name not already in
+            the GeoGraph.
+        """
+        if not set(class_list).issubset(self.df["class_label"].unique()):
+            raise ValueError("`class_list` must only contain valid class names.")
+        self.df.loc[self.df["class_label"].isin(class_list), "class_label"] = new_name
+
     def add_habitat(
         self,
         name: str,
-        valid_classes: List[int],
+        valid_classes: List[Union[str, int]],
+        barrier_classes: Optional[List[Union[str, int]]] = None,
         max_travel_distance: float = 0.0,
         add_distance: bool = False,
+        add_component_edges: bool = False,
     ) -> None:
         """
-        Create habitat graph and store it in habitats dictionary.
+        Create HabitatGeoGraph object and store it in habitats dictionary.
 
         Creates a habitat subgraph of the main GeoGraph that only contains edges
         between nodes in `valid_classes` as long as they are less than
         `max_travel_distance` apart. All nodes which are not in `valid_classes`
-        are not in the resulting habitat graph.
+        are not in the resulting habitat graph. This graph is then stored as its
+        own HabitatGeoGraph object with all meta information.
 
         Args:
             name (str): The name of the habitat.
-            valid_classes (List[int]): A list of integer class labels which make
-            up the habitat.
+            valid_classes (List): A list of class labels which make up the habitat.
+            barrier_classes (List): Defaults to None.
             max_travel_distance (float): The maximum distance the animal(s) in
             the habitat can travel through non-habitat areas. The habitat graph
             will contain edges between any two nodes that have a class label in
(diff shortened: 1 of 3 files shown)


src/models/geograph.py
```diff
@@ -488,39 +492,70 @@ def merge_nodes(
...
+            raise ValueError("`class_list` must only contain valid class names.")
+        self.df.loc[self.df["class_label"].isin(class_list), "class_label"] = new_name
+
     def add_habitat(
         self,
         name: str,
-        valid_classes: List[int],
+        valid_classes: List[Union[str, int]],
+        barrier_classes: Optional[List[Union[str, int]]] = None,
         max_travel_distance: float = 0.0,
```
rdnfn, line 520: Nice to have this now 🚀

src/models/geograph.py
```diff
@@ -576,36 +607,60 @@ def add_habitat(
...
-        'geometry' being the union of all individual polygons making up a
-        component.
-
-        Warning: this method can only be passed graphs which are a subgraph of
-        this class's main `graph` attribute, such as the habitat subgraph or the
-        main graph itself. Warning: this method is very slow if the graph
-        consists mostly of one big component, because taking the union is
-        expensive.
+        self, calc_polygons: bool = True, add_distance_edges: bool = False
+    ) -> ComponentGeoGraph:
```
rdnfn, line 652: Good to be able to turn off the calc polygons 👍

src/models/geograph.py
```diff
@@ -614,21 +669,317 @@ def get_graph_components(
...
+            comp_geograph = ComponentGeoGraph(
+                components, gdf, add_distance_edges=add_distance_edges
+            )
+        else:
+            comp_geograph = ComponentGeoGraph(components)
+        if not hasattr(self, "components"):
+            self.components = comp_geograph
+        return comp_geograph
+
+    def get_metric(self, name: str) -> metrics.Metric:
```
rdnfn, line 701: Just for my understanding: so if I want to get all available metrics I could just use the line below?
```python
all_metrics_of_graph = [graph.get_metric(name) for name in metrics.METRICS_DICT]
```

src/models/geograph.py
```diff
@@ -614,21 +669,317 @@ def get_graph_components(
...
+                data = pickle.load(file)
+        self.df = data["dataframe"]
+        self.name = data["name"]
+        self.valid_classes = data["valid_classes"]
+        self.barrier_classes = data["barrier_classes"]
+        self.max_travel_distance = data["max_travel_distance"]
+        self.add_distance = data["add_distance"]
+
+    def save_habitat(self, save_path: pathlib.Path) -> None:
+        """
```
rdnfn, line 851: For my understanding: how is this different from the main save method of the `GeoGraph` class?

src/models/metrics.py
```diff
@@ -0,0 +1,137 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any, Optional
+
+import networkx as nx
+import numpy as np
+
+if TYPE_CHECKING:
+    from src.models import geograph
```
rdnfn, line 11: Nice, didn't know there is a variable like this. Should start to use this as well!

src/models/geograph.py
```diff
@@ -614,21 +669,317 @@ def get_graph_components(
...
+        self.graph: nx.Graph = nx.Graph()
+        self.components_list: List[set] = components_list
+        if df is not None:
+            self.has_distance_edges: bool = add_distance_edges
+            self.df = self._load_from_dataframe(df, add_distance_edges)
+        else:
+            self.has_df = False
+            self.has_distance_edges = False
+            self.graph = nx.empty_graph(len(self.components_list))
+        self.metrics: Dict[str, Metric] = {}
```
rdnfn, line 931: For my understanding: this would then just be a number of nodes without any additional info?

src/models/metrics.py
```diff
@@ -0,0 +1,137 @@
...
+        value=np.sum(geo_graph.df.area.values),
+        name="total_area",
+        description="The total area of all the patches in the graph.",
+        variant="conventional",
+    )
+
+
+def _avg_component_area(geo_graph: geograph.GeoGraph) -> Metric:
+    comp_geograph = geo_graph.components
+    if not comp_geograph.has_df:
```
rdnfn, line 81: Question for usage: does a `GeoGraph` instance by default have this attribute, with a `ComponentGeoGraph` with df? Or how can I make sure that this metric can be computed for a GeoGraph?

src/models/metrics.py
```diff
@@ -0,0 +1,137 @@
...
+        value=np.mean(comp_geograph.df.area.values),
+        name="avg_component_area",
+        description="The average area of the components in the graph",
+        variant="component",
+    )
+
+
+def _avg_component_isolation(geo_graph: geograph.GeoGraph) -> Metric:
+    """Calculate the average distance to the next-nearest component."""
+    comp_geograph = geo_graph.components
```
rdnfn, line 95: Great that we can compute this, at least for smaller habitats.

Again usage-question: how to I make sure that a `GeoGraph` instance has all the relevant attributes to be able to compute this?

rdnfn (approved):
Awesome, nice work! Great to have the full metrics functionality for habitats as well!

My questions are mainly about using it: how to ensure the `GeoGraph` and its habitats have all the attributes necessary to compute all component metrics.

Otherwise, happy to merge as is! 🚀

src/models/geograph.py
rdnfn: Does this actually go into the computation yet, or is this just for future-proofing?

src/models/metrics.py
rdnfn: Also: how hard would it be to extend this to just regular patch isolation as well?

src/models/geograph.py
```diff
@@ -576,36 +607,60 @@ def add_habitat(
...
+        """Return a GeoGraph with the connected components of this graph.
+
+        This method determines the individual disconnected graph components that
+        make up the graph of the GeoGraph object. It returns a ComponentGeoGraph
+        object such that each row of the GeoDataFrame and each node in the
+        networkx.Graph correspond to a connected component in the main graph,
+        and the polygons in the dataframe are the union of all individual
+        polygons making up a component in the main graph.
+
+        Warning: this method is very slow if `calc_polygons=True` and the graph
```
Croydon-Brixton, line 662: Great to put this warning! Thank you!

src/models/geograph.py
```diff
@@ -614,21 +669,317 @@ def get_graph_components(
...
+            habitatwith edge weights that are the distance between neighbouring
+            components. Can be computationally expensive. Defaults to False.
+
+        Raises:
+            ValueError: If `data` is of an unknown type, or if `data` is a file
+            path of an invalid suffix.
+        """
+        # pylint: disable=super-init-not-called
+        if isinstance(data, gpd.GeoDataFrame):
+            # Note that index is not reset, so it contains the loc indices of the
```
Croydon-Brixton, line 782: 👍

src/models/geograph.py
Croydon-Brixton: I guess the only difference is the `data = { ... } ` part?

src/models/geograph.py
```diff
@@ -614,21 +669,317 @@ def get_graph_components(
...
+            to add edges between every pair of nodes, with the distance between
+            their corresponding polygons as an edge attribute. Defaults to False.
+        """
+        # pylint: disable=super-init-not-called
+        self.has_df: bool = True
+        self.graph: nx.Graph = nx.Graph()
+        self.components_list: List[set] = components_list
+        if df is not None:
+            self.has_distance_edges: bool = add_distance_edges
+            self.df = self._load_from_dataframe(df, add_distance_edges)
```
Croydon-Brixton, line 926: I'm slightly confused here - `_load_from_dataframe` has the signature `self, df, tolerance` and it seems here it is passed `self, df, add_distance_edges` where `add_distance_edges` is a boolean rather than a float tolerance? 

Did you want to pass a tolerance (or nothing, since you write tolerance is redundant) here instead, or where do I go wrong?

src/models/geograph.py
```diff
@@ -614,21 +669,317 @@ def get_graph_components(
...
+        else:
+            self.has_df = False
+            self.has_distance_edges = False
+            self.graph = nx.empty_graph(len(self.components_list))
+        self.metrics: Dict[str, Metric] = {}
+
+    def _load_from_dataframe(
+        self,
+        df: gpd.GeoDataFrame,
+        tolerance: float = 0.0,
```
Croydon-Brixton, line 936: What's the best-practice in this case actually? Is it to keep redundant arguments running (e.g. the way you did or with kwargs) or is it to remove them? I don't know but would be curious to learn

src/models/geograph.py
herbiebradley: It's just for future proofing at the moment.

src/models/geograph.py
```diff
@@ -614,21 +669,317 @@ def get_graph_components(
...
+            tolerance (float, optional): Redundant argument to ensure interface
+            is consistent with parent class. Defaults to 0.0.
+
+        Returns:
+            gpd.GeoDataFrame: A processed GeoDataFrame containing polygon objects.
+        """
+        # Reset index to ensure consistent indices
+        df = df.reset_index(drop=True)
+        # Using this list and iterating through it is slightly faster than
+        # iterating through df due to the dataframe overhead
```
Croydon-Brixton, line 959: Did you benchmark with `df.geometry.values` (which gives you a geometry array object, that has references to the underlying C objects? How does this compare? 
(Feel free to address this after the report and just open an issue - I think before the report submission performance is not priority)

src/models/metrics.py
```diff
@@ -0,0 +1,137 @@
...
+    from src.models import geograph
+
+
+# define a metric dataclass with < <= => > == comparisons that work as you would
+# expect intuitively
+@dataclass()
+class Metric:
+    """Class to represent a metric for a GeoGraph, with associated metadata."""
+
+    value: Any  # No good Numpy type hints
```
Croydon-Brixton, line 20: Sadly 😢 I think ppl are working on it though

src/models/geograph.py
herbiebradley: The problem is that `avg_component_isolation` doesn't work unless you have edges with distance between the components in the `ComponentsGeoGraph` of the `GeoGraph` you call the metric on, and `avg_component_area` doesn't work when the `ComponentsGeoGraph` doesn't have a dataframe with polygons, which is the default for the main `GeoGraph`. So if you do the above then you will get an error - any ideas on how to resolve this?

src/models/geograph.py
```diff
@@ -614,21 +669,317 @@ def get_graph_components(
...
+            self.graph.nodes[node]["bounds"] = polygon.bounds
+
+        # Add edge attributes if necessary
+        if self.has_distance_edges:
+            for u, v, attrs in tqdm(
+                self.graph.edges.data(data=True),
+                desc="Calculating edge weights",
+                total=len(self.graph.edges),
+            ):
+                attrs["distance"] = geom[u].distance(geom[v])
```
Croydon-Brixton, line 982: That loop and distance calculation is probably the very slow computation you mentioned right?

src/models/geograph.py
herbiebradley: This also saves all the metadata for the habitat, such as the list of valid classes & barrier classes, the `max_travel_distance`, etc. And the loading function loads all that stuff as well.

src/models/geograph.py
herbiebradley: Yeah, I thought it best to create the graph so that users can at least add edges and node attributes if they want.

src/models/geograph.py
Croydon-Brixton: Maybe a property like `available_metrics` or so would be nice? This way a user could check what possible arguments are allowed.

Croydon-Brixton (approved):
Looks good to me as well! Mainly added a few questions for my understadning. Great job @herbiebradley  🚀

src/models/metrics.py
herbiebradley: By default in the `GeoGraph` init we call `self.components = self.get_graph_components(calc_polygons=False)`, which means the resulting `ComponentsGeoGraph` does not have a df so this metric will not work there. You can calculate the df and store it in `components` with `geograph.components = geograph.get_graph_components(calc_polygons=True)`. However, by default the `HabitatGeoGraph` does get the components along with the component df in its init function, so this metric will work fine for those.

src/models/metrics.py
herbiebradley: You can do `geograph.components = geograph.get_graph_components(calc_polygons=True, add_distance_edges=True)`, to calculate the distance edges and make this metric work, but as warned this will be very expensive for graphs with more than about 100 components.

herbiebradley: For patch isolation I think the approach I've been taking to calculate distances between everything is too inefficient, so I think you would have to change the idea of the metric from being distance to next-nearest patch to being average distance between neighbouring patches.

src/models/geograph.py
```diff
@@ -7,25 +7,29 @@
...
 import rasterio
 import shapely
 from shapely.prepared import prep
 from src.data_loading import rasterio_utils
-from src.models import binary_graph_operations
+from src.models import binary_graph_operations, metrics
+from src.models.metrics import Metric
 from tqdm import tqdm
 
+pd.options.mode.chained_assignment = None  # default='warn'
```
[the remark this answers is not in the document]
herbiebradley: This is because in the `add_habitat` function, we create a new habitat by slicing the main df to get the rows which correspond to the habitat. Further operations on the habitat df then give a Pandas warning about chained assignment, which is explained here in the first answer: https://stackoverflow.com/questions/20625582/how-to-deal-with-settingwithcopywarning-in-pandas, but for our use case we don't care about it so I disabled the warning.

src/models/geograph.py
herbiebradley: I think it's to create the redundant arguments so that if people accidentally copy a function call from the superclass and apply it to the child class, it all still works. See https://github.com/python/mypy/issues/1237 https://en.wikipedia.org/wiki/Liskov_substitution_principle

src/models/geograph.py
herbiebradley: Yeah I was thinking about this, I'll test it out before I merge the PR.

src/models/geograph.py
```diff
@@ -530,21 +565,17 @@ def add_habitat(
         idx_dict: Dict[int, int] = dict(zip(range(len(self.df)), self.df.index.values))
         # Get dicts of polygons and buff polygons to avoid repeatedly querying
         # the dataframe. These dicts accept loc indexes
-        polygons: Dict[int, shapely.Polygon] = self.df.geometry.to_dict()
+        polygons: Dict[int, shapely.Polygon] = self.df["geometry"].to_dict()
```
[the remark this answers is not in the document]
herbiebradley: Preferred style - there are quite a few things you can't do with dot notation in pandas, so we would naturally end up with some with dot notation and some with dict notation if we used `.geometry`. Therefore by changing to `["geometry"]` we get a unified style which can also be a bit easier to read.

src/models/geograph.py
herbiebradley: See https://github.com/ai4er-cdt/gtc-biodiversity/pull/42#discussion_r592249401

herbiebradley: Ah yes, my bad - I didn't mean to pass a tolerance here, just df.

src/models/geograph.py
```diff
@@ -488,39 +492,70 @@ def merge_nodes(
...
+            class_list (List): The list of names of class labels to combine.
+            Every name in the list must be in the GeoGraph.
+
+        Raises:
+            ValueError: If `class_list` contains a class name not already in
+            the GeoGraph.
+        """
+        if not set(class_list).issubset(self.df["class_label"].unique()):
+            raise ValueError("`class_list` must only contain valid class names.")
+        self.df.loc[self.df["class_label"].isin(class_list), "class_label"] = new_name
```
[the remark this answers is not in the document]
herbiebradley: I should be able to find these cases and call the very convenient `merge_nodes` function to deal with it :D, I'll add it to the PR tomorrow probably.

src/models/geograph.py
herbiebradley: One thing you can do is 
```python
for name in metrics.METRICS_DICT:
    try:
        all_metrics.append(graph.get_metric(name))
    except ValueError:
        continue
```
which would not give any errors and should return all the metrics it is possible to calculate.

src/models/geograph.py
```diff
@@ -168,29 +164,26 @@ def __init__(
...
 
         print(
             f"Graph successfully loaded with {self.graph.number_of_nodes()} nodes",
             f"and {self.graph.number_of_edges()} edges.",
         )
 
+    def __eq__(self, o: object) -> bool:
+        if not isinstance(o, GeoGraph):
+            return False
+        return nx.is_isomorphic(self.graph, o.graph)
```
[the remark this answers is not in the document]
herbiebradley: Yeah I just realised it's exponential in the number of vertices - for some reason that function is still very fast when comparing graphs which have quite a few differences though, but impossibly slow for graphs which are the same. I've changed it to `nx.fast_could_be_isomorphic` which takes about half a second to compare even for graphs which are equal. It's only an approximation of course but I think it will be quite accurate for the differences usually found between GeoGraphs.

src/models/geograph.py
herbiebradley: Yeah that's that one

src/models/metrics.py
rdnfn: Also also: should we set this to zero if there is only a single component? I think this would make sense, as we would want this metric to go down as more components become connected, and in the limit (i.e. all components are connected) it should be the lowest = 0.

herbiebradley: Good idea, done in f5b80c0

src/models/geograph.py
rdnfn: I think given the computational cost to create some of these metrics scales so poorly we should keep the GeoGraph defaults scalable (i.e. not compute all the component edges, df etc.). The most important immidiate thing will be to add good warnings (and docstrings with examples) that let the user know how to add the relevant info to compute a more sophisticated metric.

In the next step, we could let the metrics themselves trigger the computation of these data structures. So e.g. if you want to compute `avg_component_area` it calls a `GeoGraph` method `get_components(datafram=True)`, that computes the components with df if the current component attribute doesn't have this yet. This would make it easier for the user, but maintain default scalability.

src/models/geograph.py
rdnfn: Yep, sounds good.

src/models/metrics.py
rdnfn: Thanks for the info, I think the best quick addition for this PR will be to add this fix 

```(geograph.components = geograph.get_graph_components(calc_polygons=True, add_distance_edges=True)```

 to the warning and docstring of the two component metric functions. Then as a next step we could extend the get components function as I suggested above.

src/models/geograph.py
herbiebradley: We can certainly make it so that the metrics functions trigger the computation of the needed data structures, but that would then make it impractical to use the code above iterating over all metrics, because it would trigger some incredibly expensive computation unless your graph is small. I personally slightly prefer allowing users to get all metrics calculatable with the defaults easily. Maybe I should add a `get_all_metrics` function which gets all the metrics that are valid for a geograph - then additional ones can be called optionally.

src/models/metrics.py
rdnfn: (Again beyond the scope of this PR) About patch and component isolation: didn't give this extensive thought but I think we can use the existing graph to make this more information:
1. only computing distances where there are edges (because if there is an edge, the minimal distance patch must has such an edge)
2. slowly searching for nearby patches (like creating a larger max_travel_distance habitat) by adding a double_max travel distance buffer and then adding the edges with distance to the graph, and doing this in incriments of the max_travel_distance, until we found an edge.
Together this should still be more efficient in most cases than the complete distance computation at the moment.

src/models/geograph.py
rdnfn: Hm yeah I think you're right, it would be nice for the user to have an easy way to get the simple metrics. Maybe we could have `get_standard_metrics` (for computationally inexpensive) and `get_extended_metrics` (for all metrics including expensive) functions?

rdnfn: Generally, I think triggering these computations is what a user would expect, if you try to get a complicated metric. But you're absolutely right, it shouldn't be the default anywhere.

herbiebradley: Alright, in b196db1 I added a `STANDARD_METRICS` list and an `ALL_METRICS` list, so you can just iterate over the standard metrics list to get all of the cheap metrics. Then I made the metrics functions compute the needed data structures automatically, with a warning.

src/models/geograph.py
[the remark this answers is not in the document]
herbiebradley: This is taking more time than I thought it would, so I'll create an issue and add it in a future PR.

src/models/geograph.py
herbiebradley: @Croydon-Brixton I did some tests and it turns out making the list is slower than just getting the `.values`, but accessing a random element in the list is ~25x faster than accessing a random element in the geometry array (which is built on top of numpy I think). So we can stick to lists, this post might explain why they are better for accessing elements: https://stackoverflow.com/questions/29281680/numpy-individual-element-access-slower-than-for-lists

herbiebradley:
All points dealt with, time to merge! 🥳

PR #47, opened 2021-03-15
Feature/graph analysis ipyleaflet visualisation

rdnfn:
Introducing `GeoGraphViewer` 🎉 ... this PR adds the `GeoGraphViewer` visualisation class and a bunch of code that enables it. As the `GeoGraphViewer` is a fairly complex widget, this PR is somewhat lengthy, apologies in advance! Thanks a lot for your review!

**Main changes:**
- All GeoGraph visualisation code now lives in `visualisation/*`. The old `models/visualisation.py` has been removed.
- Added `GeoGraphViewer` class (in `visualisation/geoviewer.py`): this allows us to visualise GeoGraphs interactively.
- Added control widgets (in `visualisation/control_widgets.py`): these make up the side control panel seen in demos. See architecture image below for a rough outline how they relate and are added to the `GeoGraphViewer`.

**Other changes:**
- Added different utils that are used by `GeoGraphViewer`. These are in 
  - `visualisation/widget_utils.py`
  - `visualisation/folium_utils.py` (old functions from `visualisation.py`, **no need to review**)
  - `visualisation/graph_utils.py` (old functions from `visualisation.py`, **no need to review**)
- Added a new test case to `data_loading/test_data.py`
- Added `visualisation/style.py` that contains constants defining the plotting style of graph, etc. in `GeoGraphViewer`.
- Added ipyleaflet and jupyter *lab* dependencies in `requirements/dev-requirements.txt`
- Added misc exploratory notebooks that helped me develop the viewer, for reviewers only `notebooks/exploratory/rdnfn-8-polesia-geographviewer-demo.ipynb` will likely be interesting, which is the demo on the Polesia data. Feel free to use it, it should work with the environment created via `make env` (if you updated it to new requirements)
- The old folium code lives on without further development in the `FoliumGeoGraphViewer` class (in `geoviewer.py`)
- Updated readme to new visualisation folder

**Architecture**
Screenshot 2021-03-15 at 20 51 38

**GeoGraphViewer demo**
Screenshot 2021-03-15 at 20 47 42


Croydon-Brixton:
Love it already before reviewing! 😉 😍

src/visualisation/geoviewer.py
```diff
@@ -0,0 +1,414 @@
...
+        super().__init__(
+            center=center,
+            zoom=zoom,
+            scroll_wheel_zoom=True,
+            crs=ipyleaflet.projections.EPSG3857,  # EPSG code for WGS84 CRS
+            zoom_snap=0.1,
+            **kwargs
+        )
+        # There seems to be no easy way to add UTM35N to ipyleaflet.Map(), hence WGS84.
+        self.gpd_crs_code = WGS84
```
herbiebradley, line 61: What does this affect? Since WGS84 is a degrees-based CRS and UTM35N is not.

src/visualisation/geoviewer.py
```diff
@@ -0,0 +1,414 @@
...
+        for layer_type, type_dict in self.layer_dict.items():
+            for layer_name in type_dict:
+                if layer_type == "maps":
+                    self.set_layer_visibility(layer_type, layer_name, "map", False)
+                elif layer_type == "graphs":
+                    self.set_layer_visibility(layer_type, layer_name, "graph", False)
+                    self.set_layer_visibility(layer_type, layer_name, "pgons", False)
+                    self.set_layer_visibility(
+                        layer_type, layer_name, "components", False
+                    )
```
herbiebradley, line 137: Does this hide all layers if we have a disconnected nodes or poorly connected nodes layer?

src/visualisation/geoviewer.py
```diff
@@ -0,0 +1,414 @@
...
+        self.layer_update()
+
+    def add_layer(self, layer: Union[dict, ipyleaflet.Layer], name=None) -> None:
```
herbiebradley, line 140: is this function used by anything? I couldn't find an explicit call anywhere.

src/visualisation/geoviewer.py
```diff
@@ -0,0 +1,414 @@
...
+                name=current_name + "_components",
+                **self.layer_style["components"]
+            )
+
+            # Creating layer for disconnected (no-edge) nodes
+            disconnected_nx_graph = nx_graph.subgraph(
+                [node for node in nx_graph.nodes() if nx_graph.degree[node] == 0]
+            )
+
+            discon_nodes, _ = graph_utils.create_node_edge_geometries(
```
herbiebradley, line 219: Can we not use `include_edges=False` here to save computing the edges?

herbiebradley: And the same applies to the call below as well.

src/visualisation/geoviewer.py
```diff
@@ -0,0 +1,414 @@
...
+
+        Args:
+            layer_type (str): type of layer (e.g. "maps","graphs")
+            layer_name (str): name of layer
+            layer_subtype (str): subtype of layer (e.g. "map","components")
+            active (bool): whether layer is activate (=visible)
+        """
+        self.layer_dict[layer_type][layer_name][layer_subtype]["active"] = active
+
+    def hidde_all_layers(self) -> None:
```
herbiebradley, line 126: minor nitpick: typo

src/visualisation/graph_utils.py
```diff
@@ -0,0 +1,50 @@
...
+
+    Returns:
+        Tuple[gpd.GeoDataFrame, gpd.GeoDataFrame]: dataframes of nodes and edges
+            respectively.
+    """
+
+    node_gdf = gpd.GeoDataFrame(columns=["id", "geometry"])
+    rep_points = graph.nodes(data="rep_point")
+    for idx, rep_point in rep_points:
+        node_gdf.loc[idx] = [idx, rep_point]
```
herbiebradley, line 35: Since `.loc` is quite slow, you should be able to speed this up quite a lot by just initialising the `node_gdf` using `rep_points`: https://www.geeksforgeeks.org/creating-a-pandas-dataframe-using-list-of-tuples/

src/visualisation/graph_utils.py
```diff
@@ -0,0 +1,50 @@
...
+    node_gdf = node_gdf.set_crs(crs)
+
+    if include_edges:
+        edge_gdf = gpd.GeoDataFrame(columns=["id", "geometry"])
+        for idx, (node_a, node_b) in enumerate(graph.edges()):
+            point_a = rep_points[node_a]
+            point_b = rep_points[node_b]
+            line = shapely.geometry.LineString([point_a, point_b])
+
+            edge_gdf.loc[idx] = [idx, line]
```
herbiebradley, line 45: And similarly here, but you'll need to build the list of tuples in the loop first then make the `edge_gdf` from it at the end.

herbiebradley (approved):
All done, looks good to me! I made some suggestions, a few are minor nitpicks and a couple more are performance improvements. Since time is tight there's no need to resolve those before the deadline if you're too busy.

src/visualisation/geoviewer.py
rdnfn: The CRS of the viewer application always is WGS84, because ipyleaflet doesn't support UTM 35N or similar natively as far as I could spot it. So any geometry is reprojected from geo_graph.crs to WGS84 before being added as a layer (graph, components, pgons, etc.). This shouldn't affect any geograph metrics as this all happens internally in the viewer. It's not optimal, but I don't think it's adds major complexity as this is only done once on loading (add_graph()).

src/visualisation/geoviewer.py
rdnfn: Good catch! Fixed it in an upcoming commit.

src/visualisation/geoviewer.py
rdnfn: It's not used internally in any widget but it overrides an existing ipyleaflet function. It enables users to add custom (non-graph) layers to the viewer and they will show up as map layers in the UI (good for showing satellite data for example)

src/visualisation/geoviewer.py
rdnfn: Yeah, good point. I implemented that argument for those two calls but then didn't use it.

src/visualisation/graph_utils.py
rdnfn: Good point! I added a performance improvement issue, so I don't forget it in the future.

src/visualisation/graph_utils.py
rdnfn: See above.

rdnfn:
Thanks a lot for the review @herbiebradley! Was able to address most points in 42eae9e, and created Issue #50 for the remaining performance improvements, to be dealt with at a later point.

PR #48, opened 2021-03-16
Feature/graph analysis merging

Croydon-Brixton:
As discussed, I'm opening this PR to the work in progress `graph-analysis-merging` branch to speed up the PR process.

I will add more documentation and an overview of this PR once it's fully ready 😉


src/models/geograph.py
```diff
@@ -189,6 +190,11 @@ def crs(self):
         """Return crs of dataframe."""
         return self.df.crs
 
+    @property
+    def bounds(self):
+        """Return bounds of entire graph"""
+        return self.df.total_bounds
```
herbiebradley, line 196: Suggestion, change this to `self.df.sindex.bounds` 
image

herbiebradley, line 196:
```suggestion
        return self.df.sindex.bounds
```

herbiebradley: Minor point, but it seems a lot faster 
image

src/models/geograph.py
```diff
@@ -698,34 +696,186 @@ def get_graph_components(
...
+    def _remove_node(self, node_id: int):
+        self._remove_nodes([node_id])
+
+    def _remove_nodes(self, node_ids: Iterable[int]):
+        # Remove node from graph (automatically removes edges)
+        self.graph.remove_nodes_from(node_ids)
+        # Remove data of node from df
+        self.df = self.df.drop(index=node_ids)
+
+    def _add_node(
```
herbiebradley, line 834: Nice, these functions are very convenient.

src/models/geograph.py
```diff
@@ -795,6 +945,9 @@ class label in `valid_classes`, as long as they are less than
                 self.barrier_classes: List[Union[str, int]] = barrier_classes
                 self.max_travel_distance: float = max_travel_distance
                 self.add_distance: bool = add_distance
+                self.metrics: Dict[str, Metric] = {}
+                self.class_metrics: Dict[str, Dict[str, Metric]] = {}
```
herbiebradley, line 949: I think we don't want `metrics` and `class_metrics` inside the `isinstance`, right? Because if loading from file we also want to initialise these.

src/utils/geopandas_utils.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+
+    # Identify the nodes that will be merged
+    mapping = identify_dfs(df, df, mode="corner")
+    mapping_graph = nx.from_dict_of_lists(mapping)
+    nodes_to_merge = [
+        list(group)
+        for group in nx.algorithms.connected_components(mapping_graph)
+        if len(group) > 1
+    ]
```
herbiebradley, line 120: Very elegant! ;)

src/models/geograph.py
Croydon-Brixton: Nice, thank you!

Croydon-Brixton:
@herbiebradley @rdnfn Branch is now ready for review. I'm aware that a detailed review today will likely not be possible, so don't feel pressured to. For the publicly deployed branch we can do that after the report deadline if needed. 

Main new features:
- *NodeMap*: An object to store node identifications between two GeoGraph objects. It's essentially a bipartite graph. This forms the basis for timestacking. It can be inverted easily.
- *GeoGraphTimeline*: The main new addition. Allows to stack several geographs in time, perform node identification between them (via `.timestack()`) and to analyse time-series of landscape, habitat, class and patch level metrics. The node identification also allows us to determine absolute and relative growth rates for nodes as well as classifying the `node-dynamics` type. While the functions for node-dynamics calculations, etc. are currently written in pure python, in my experiements they always exectued within under a second on graphs with ~2000 nodes. The major performance bottleneck are still edge computations when building up habitats or for getting e.g. `avg_component_isolation` across time. I think several of these might be alleviated sligthly by casting the performance relevant code into `cython` or `numba`-izing what can be transformed with numba. But that's definitely something for after the report 😉 
- *Connectivity=8* now supported: We can fix the GDAL problem via our node identification on a connectivity=4 polygonised raster image. Connectivity=8 (diagonal patch connectivity) is important becauses it is the usual definition in rasterised landscape ecology. So using connectivity=8 is required to compare our metrics to Fragstats/pylandstats. Connectivity=4 is still supported as before.

Other changes:
- *TimedGeoGraph*: A GeoGraph with a `self.time` attribute. Inherits all else from GeoGraph.
- Handy wrappers for adding and removing nodes from GeoGraphs
- Minor bugfixes (e.g. crs handling of GeoGraph, ...)
- Demo notebooks that show how to plot and perform the time-series analysis. I will still clean those up a bit today.

src/models/geograph.py
Croydon-Brixton: Sorry , I am not sure I fully understand  😉  Do you mean to move them to the `_load_from` methods such that they can be loaded from memory, if applicable? If yes, I think that'd be a good idea

herbiebradley: I think that there are situations where we might not want to load metrics, so let's leave that - but at the moment if we load from file the `metrics` and `class_metrics` dicts are not initialised so if someone calls `get_metric` after loading from file I think it won't work?

src/models/geotimeline.py
```diff
@@ -1,6 +1,522 @@
...
+
+# type alias
+TimeStamp = Union[int, datetime.datetime]
+
+
+class NotCachedError(Exception):
+    """Basic exception for values which were not yet cached."""
+
+
+class TimedGeoGraph(GeoGraph):
```
herbiebradley, line 23: I can see why you would want a time attribute, but what made you choose this way instead of just adding an attribute to `GeoGraph`?

src/models/geotimeline.py
```diff
@@ -1,6 +1,522 @@
...
+    def _sort_by_time(self, reverse: bool = False) -> None:
+        """
+        Sort the graphs in GeoGraphTimeline accroding to timestamp from earlier to later
+
+        Args:
+            reverse (bool, optional): If False, start with the earliest date.
+                If True, sort starting with the latest date. Defaults to False.
+        """
+        self._graphs = {
+            time: self._graphs[time] for time in sorted(self._graphs, reverse=reverse)
```
herbiebradley, line 120: Nice to take advantage of the Python 3.7 feature to have ordered dicts by default!

src/models/geotimeline.py
```diff
@@ -1,6 +1,522 @@
...
+        metrics = [
+            graph.get_metric(name=name, class_value=class_value) for graph in self
+        ]
+
+        # Set up metadata
+        attrs = {"description": metrics[0].description, "unit": metrics[0].unit}
+        if class_value:
+            attrs["class_label"] = class_value
+
+        metric_timeseries = xr.DataArray(
```
herbiebradley, line 243: I'm curious - what makes the `DataArray` more convenient here?

src/models/geotimeline.py
```diff
@@ -1,6 +1,522 @@
...
+        # Initialize empty node map cache dictionary
+        self._node_map_cache: Dict[(TimeStamp, TimeStamp), NodeMap] = dict()
+
+    @property
+    def times(self) -> List[TimeStamp]:
+        """Return list of valid time stamps for this GeoGraphTimeline"""
+        return list(self._graphs.keys())
+
+    @property
+    def graphs(self) -> List[GeoGraph]:
```
herbiebradley, line 92:
```suggestion
    def graphs(self) -> Dict[TimeStamp, GeoGraph]:
```

src/models/metrics.py
```diff
@@ -126,30 +465,68 @@ def _avg_component_isolation(geo_graph: geograph.GeoGraph) -> Metric:
...
+            try:
+                return LANDSCAPE_METRICS_DICT[name](geo_graph, **metric_kwargs)
+            except KeyError as key_error:
+                return COMPONENT_METRICS_DICT[name](geo_graph, **metric_kwargs)
+        except KeyError as key_error:
+            raise ValueError(
+                f"Argument `{name}` is not a valid landscape/component metric.\n"
+                "Available landscape metrics are: "
+                f"\n{list(LANDSCAPE_METRICS_DICT.keys())}.\n"
+                "Available landscape metrics are: "
```
herbiebradley, line 518:
```suggestion
                "Available component metrics are: "
```

src/models/geotimeline.py
```diff
@@ -1,6 +1,522 @@
...
+            nodemap (NodeMap): A node mapping between two GeoGraphs. Nodes in the source
+                graph of the mapping are referred to as `ancestors`, nodes in the target
+                graph are referred to as `children`.
+
+        Returns:
+            str: The identified node dynamics. One of `birth`, `split`,
+                `morphed/unchanged`, `complex`, `merged`.
+        """
+        # TODO: numba typed dict for speedup
+        ancestors = (~nodemap).mapping[child_index]
```
herbiebradley, line 468: Love the `__inverse__`!

src/models/geotimeline.py
```diff
@@ -1,6 +1,522 @@
...
+            node_map = self.node_map_cache(prior_time, time)
+
+            # Helper function to calculate dynamics type via a single map call
+            dynamics_oracle = (
+                lambda child_index: GeoGraphTimeline._calculate_node_dynamics(
+                    child_index, node_map
+                )
+            )
+
+            dynamics_type = self[time].df.index.map(dynamics_oracle)
```
herbiebradley, line 424: Does this return a `Series` (so it has loc indexes) or just a List (so it has iloc indexes)?

src/models/geotimeline.py
```diff
@@ -1,6 +1,522 @@
...
+
+            self[time].df["relative_growth"] = rel_growth
+            self[time].df["absolute_growth"] = abs_growth
+            self[time].df["node_dynamic"] = dynamics_type
+
+            # Detect dynamics type of unchanged/morphed by comparing to the
+            # relative growth
+            grew = (dynamics_type == "morphed/unchanged") & (rel_growth > 0)
+            shrank = (dynamics_type == "morphed/unchanged") & (rel_growth < 0)
+            unchanged = (dynamics_type == "morphed/unchanged") & (rel_growth == 0)
```
herbiebradley, line 435: Nice use of bitwise and between boolean arrays!

src/models/geograph.py
herbiebradley: Resolved in a2050ef

herbiebradley (dismissed):
All done! Looks good to merge, I just made a commit adding a docstring to `_add_node` in `geograph.py`, and improving some type hints.

herbiebradley (dismissed):
LGTM 🚀

src/models/geotimeline.py
Croydon-Brixton: This returns a Series 😉

src/models/geotimeline.py
Croydon-Brixton: The multidimensionality and the fact that we can add the nice description that you wrote for the metrics as well as units to it 😉

src/models/geotimeline.py
Croydon-Brixton: True, at the time of creation I wasn't sure if I was going to exted a timed geograph more, I guess I just wanted to have a play with polymorphism 😆  
Not really needed here I agree, can get rid of it in the next update!

Croydon-Brixton:
> All done! Looks good to merge, I just made a commit adding a docstring to `_add_node` in `geograph.py`, and improving some type hints.

Thanks a lot, that's excellent!

herbiebradley (approved):
Resolved merge conflicts in a hacky way, will sort it out later. Time to merge ;)

PR #49, opened 2021-03-16
fix: changed font so serif without tex. New subplot label option

sdat2:
I realised that the font option would default to sans-serif when latex wasn't available.
It also makes more sense to put the subplot label above the top left of the panel when you plot an image,
rather than within the panel itself, so I added that option if anyone needs it.


Croydon-Brixton (approved):
Looks good to me 😉

PR #51, opened 2021-03-18
Feature/landcover classification unet

herbiebradley:
This should be basically good enough to merge - any last commits can be added here.


herbiebradley (commented):
I've checked it over and renamed `model.py` to `unet_model.py`, so we should be all good to merge now.

Croydon-Brixton (approved):
Awesome, thanks a lot! Yes I think it's good to merge

PR #53, opened 2021-03-18
Feature/landcover classification

herbiebradley:
This will merge all the XGBoost and UNet classification code into master. If there is any final tidying up to do with this, best to do it in this PR.


sdat2:
ah ok, so I think I just need to go in and change my files, will do that now.

sdat2 (approved):
Let's merge :) .

PR #59, opened 2021-03-23
Docs/sphinx html docs

rdnfn:
This is a quick PR that add the up to date documentation configuration to the main branch.

Docs are available here: https://geograph.readthedocs.io/en/latest/

**Main changes:**
- Update rst files (that define which modules are shown in docs) to new repo structure
- Create doc requirements (run by readthedocs)

**Other changes:**
- Updated docstrings in some packages to comply with requirements for sphinx
- Added documentation to main readme

EDIT: it looks like quite a few files changed, but these are mainly doc specific rst files.


herbiebradley (approved):
Looks good! One question: if we update the `main` branch after this PR is merged then will the docs update automatically? Or is there a way to do this in the future, if not?

rdnfn:
> Looks good! One question: if we update the `main` branch after this PR is merged then will the docs update automatically? Or is there a way to do this in the future, if not?

Thanks! At the moment I trigger the docs builds manually, but in the future I should be able to add a webhook for it.

.readthedocs.yml
```diff
@@ -0,0 +1,25 @@
...
+
+# Optionally build your docs in additional formats such as PDF
+formats:
+   - pdf
+
+# Optionally set the version of Python and requirements required to build your docs
+python:
+   version: 3.8
+   install:
+   - requirements: requirements/dev-requirements.txt
```
Croydon-Brixton, line 25:
```suggestion
   - requirements: requirements/requirements.txt
```

Would update this to the runtime dependencies rather than the requirements for developers

docs/conf.py
```diff
@@ -60,15 +60,19 @@
...
 
+html_logo = "images/geograph_logo.png"
+
+html_theme_options = {
+    'style_nav_header_background': '#a0a0a0',
+    'logo_only': True,
+    'display_version': True,
+}
+
+html_favicon = 'images/geograph_logo_small.png'
```
Croydon-Brixton, line 71: Sweet! 🤩

requirements/doc-requirements.txt
```diff
@@ -0,0 +1,78 @@
+# NOTE: These requirements are used for developing code on the repo.
```
Croydon-Brixton, line 1: Are those essentially `requirements` + `sphinx`?

Croydon-Brixton (approved):
Looks good to me too! 👍

requirements/doc-requirements.txt
rdnfn: Yes, these are the dev-requirements plus sphinx and minus the weird local path dependent package installation.

Would there be a cleaner way to do this? Maybe we should just remove the weird path dependency in dev-requirements (by changing dir in make) and add sphinx to dev? In development we might also want to be able to build docs locally?

.readthedocs.yml
rdnfn: Good catch! This should actually be doc-requirements

requirements/doc-requirements.txt
Croydon-Brixton: Not quite sure bc. I'm not as familiar with Sphinx - are those run anywhere? 
I think in any case there are several `dev-requirements` that are not needed anymore. I did not update the `dev-requirements` for now, but the `requirements` is the minimal list of requirements that are needed to have GeoGraph run

.readthedocs.yml
rdnfn: Done in ca34f1c

Croydon-Brixton (approved):
Re-approved 😉

requirements/doc-requirements.txt
Croydon-Brixton: I guess sphinx probably needs even fewer as it doesn't even execute the geograph code?

rdnfn: Cool! Let's return to this when we define all the requirements for the package itself. We can then do this in a separate PR.

Croydon-Brixton: Sounds good to me!

PR #60, opened 2021-03-23
Feature/graph analysis ipyleaflet visualisation

rdnfn:
Yet another quick PR... only two small changes that are geographviewer-related:
- Added `small_screen` argument to geographviewer that puts metrics in its own tabs (this helps when zooming in for better readability in presentations or the report)
- Added demo notebook for the geographviewer (currently still based on Jasmin data path)


herbiebradley (approved):
LGTM - definitely a good idea to have the widget scale down.

PR #63, opened 2021-03-25
v0.01 of the publishable package

herbiebradley:
This PR contains the `setup.py` necessary to define the package, along with an upload command. In the future when we want to upload all we need to do is increment the version in `_version.py`, then run `python setup.py upload`. It will ask for a PyPI username and password, then upload the package.

This PR also renames `src` to `geograph`, and moves a couple of constants to `geograph/demo/binder_constants.py`.


PYPI_README.md
```diff
@@ -0,0 +1,49 @@
...
+cd gtc-biodiversity
+make env
+```
+
+Done, you're ready to go!
+
+## 3. Requirements
+
+GeoGraph is written in Python 3.8 and builds on [NetworkX](https://github.com/NetworkX/NetworkX), [ipyleaflet](https://github.com/jupyter-widgets/ipyleaflet) and many more packages. See the [requirements directory](./requirements) for a full list of dependencies.
+
```
rdnfn, line 45:
```suggestion
GeoGraph is written in Python 3.8 and builds on [NetworkX](https://github.com/NetworkX/NetworkX), [ipyleaflet](https://github.com/jupyter-widgets/ipyleaflet) and many more packages. See the [requirements directory](https://github.com/ai4er-cdt/gtc-biodiversity/tree/main/requirements) for a full list of dependencies.
```
Since it's not on GitHub this link needs to be absolute.

geograph/binary_graph_operations.py
```diff
@@ -7,11 +7,11 @@
...
 from shapely.geometry.polygon import Polygon
 
-import src.utils.geopandas_utils as gpd_utils
-from src.utils.polygon_utils import EMPTY_POLYGON, collapse_empty_polygon
+import geograph.utils.geopandas_utils as gpd_utils
+from geograph.utils.polygon_utils import EMPTY_POLYGON, collapse_empty_polygon
 
 if TYPE_CHECKING:
-    from src import geograph
+    import geograph
```
rdnfn, line 14: I don't think this is the same thing: one imports the package, the the other module `geograph.geograph`.

geograph/geotimeline.py
```diff
@@ -9,8 +9,8 @@
 import pandas as pd
 import xarray as xr
 
-from src.binary_graph_operations import NodeMap, identify_graphs
-from src.geograph import GeoGraph
+from geograph import GeoGraph
```
rdnfn, line 12: See comment above (module vs. package), unless you have elevated GeoGraph to be directly importable?

geograph/metrics.py
```diff
@@ -8,7 +8,7 @@
 import numpy as np
 
 if TYPE_CHECKING:
-    from src import geograph
+    import geograph
```
rdnfn, line 11: See above.

geograph/tests/utils.py
```diff
@@ -106,7 +105,7 @@ def polygonise_sub_array(
 
 
 def plot_identified_nodes(
-    node: dict, other_graph: GeoGraph, identified_nodes: Iterable[int]
+    node: dict, other_graph: geograph.GeoGraph, identified_nodes: Iterable[int]
```
rdnfn, line 108: Same as above.

geograph/visualisation/folium_utils.py
```diff
@@ -5,11 +5,10 @@
 from typing import Callable, List, Optional, Tuple
 
 import folium
+import geograph
```
rdnfn, line 8: Same as above. (I may very well be completely wrong here)

setup.py
```diff
@@ -1,12 +1,146 @@
...
+    "folium",
+    "ipyleaflet",
+    "tqdm",
+    "geopandas",
+    "shapely",
+    "rtree",
+    "rasterio",
+    "xarray",
+    "networkx",
+]
```
rdnfn, line 41: The versions of these dependencies are not pinned at the moment right? That's probably something to do for future releases -> v.0.0.2 and onwards...

rdnfn (commented):
Awesome to have the package ready now! Really only had one question: whether the `GeoGraph` class is directly accessible from the package or from `geograph.geograph`. Dependent on the answer most of my comments can be safely ignored. Great work!

geograph/geotimeline.py
herbiebradley: Yes, I have - check `__init__.py` in the top level. I felt it was better to be able to do `from geograph import GeoGraph` or `import geograph` then `geograph.GeoGraph` rather than adding the extra `geograph` in the middle.

rdnfn: Nice, yep that's definitely cleaner!

rdnfn (dismissed):
Looks good to me then, ready to merge! 🚀

PYPI_README.md
herbiebradley: Done. I'll also make a note to rewrite this readme for the 0.0.2 version since at the moment it gives the instructions for cloning from github.

setup.py
herbiebradley: Yes, but it is usually not considered best practice to pin exact dependencies in `setup.py`, since it is intended as a list of the minimal requirements that are to be as flexible as possible: https://stackoverflow.com/a/44938662. Pinning dependencies to ranges would be fine of course, but at the moment we don't know what the ranges should be.

rdnfn (approved):
Nice, same as before!

Croydon-Brixton (dismissed):
Looks good to me - I added the refactorings in the notebooks and tested them. Everything works fine, good to merge from my side 😉

Croydon-Brixton (approved):
Re-approved 😉

PR #74, opened 2021-05-19
add zenodo badge

sdat2:


herbiebradley:
Good to see it's working now, when I first made the DOI it hadn't updated the zenodo DB.

herbiebradley (approved):

PR #75, opened 2021-05-28
docs: add WCMC and BTO presentation tutorial

Croydon-Brixton:
Changes:
- Adds an extra tutorial notebook (6) which was presented at the WCMC and BTO


herbiebradley (approved):
Looks good!

PR #80, opened 2022-07-13
[hotfix] change default requirements to production requirements

Croydon-Brixton:
Todo:

- [x] Fix dependency versions in requirements.txt
- [x] Clean up requirements folder (should only have environment.yaml and production reqirements)
- [x] Verify correct installation with fixed dependencies


herbiebradley (approved):

herbiebradley:
Feel free to merge - we're figuring out the right dependencies with `rasterio` in #81. I also want to include `matplotlib` and `seaborn` in the `requirements.txt`, but probably simpler to include that in the above PR rather than this one.

Croydon-Brixton:
Fixed binder and requirements jungle 😉 

✅ Requirement pins were tested successfully, both on MacOS & Ubuntu 
✅ Resolves #83

rdnfn (approved):
Thanks Simon for all this work! Very good to have all of these requirements updated. I know this stuff takes a lot of work, so your contribution is much appreciated!

